PR #3, opened 2017-11-28
More thorough and compact

makepost:
Nice generator. For [acme-commander](https://github.com/makepost/acme-commander), some changes were needed. See commit descriptions.


sammydre:
I've read through the changes and they look good. I need to test it with my application next; if that works well then I'll merge this.

sammydre:
This doesn't appear to be quite right yet, at least around the change to the static side of classes / constructors.

Let's take Gtk.Paned as an example. Here's what we have after your changes on the static side of the class:

```
static name: string
static new (config?: Paned_ConstructProps): Paned
constructor (config?: Paned_ConstructProps)
```

Here's what used to be there:

```
export interface Paned_Static {
    name: string
    new (config?: Paned_ConstructProps): Paned
}
export declare class Paned_Static {
    new(orientation: Orientation): Paned
}
export declare var Paned: Paned_Static
```

Note the difference: the "new(orientation: Orientation)" method. I believe this is correct; see e.g. the docs here: http://devdocs.baznga.org/gtk30~3.22.12/gtk.paned#constructor-new

This seems fixable with your approach, I've just not quite tracked down where this regression was introduced in your changes. As this breaks code, I think this is a blocker for merging this PR.

More generally, do you think moving from interface+var+static => class is worth it? I chose the former carefully; it avoids problems when inheriting (due to explicitly breaking up the instance and static sides of classes; there can be no problems with the static side clashing when inheriting).

makepost:
Without the move, can't use typings in plain JavaScript, see e859bde description.

Does any actual app depending on `ts-for-gjs` make use of overloaded `new`? I can get it back by reverting 582a501 and e476e7b, but that would lose the file size (parsing speed) benefit of inheritance.

[Skipping `new` overloads](https://github.com/makepost/ts-for-gjs/commit/e476e7bdb0afc63eb9400c2918950b5e4c12ca0e#diff-6427dc610ec40c788cba04a9c39d6dcfR1028) prevents conflicts that appear because static members are inherited. For example, it allows to extend GObject, which would otherwise have:

```typescript
static new (config?: Object_ConstructProps): Object

// Breaks idiomatic JS-style inheritance.
static new(object_type: number, names: string[], values: Value[]): Object
```

Overloading of `new` is useful in C, less so in JS where construction with object literal props is available:

```javascript
new Paned({ orientation: Orientation.VERTICAL });

new Paned({
  // ... other props ...
  orientation: Orientation.VERTICAL,
  // ... other props ...
});
```

Unrelated, fixed non-overloaded constructors in 8f12ae0.

sammydre:
To respond to your earlier question:

> Does any actual app depending on ts-for-gjs make use of overloaded new? 

The answer to this is "yes." The main user of ts-for-gjs today is an internal app (not yet released, nor open sourced) at my place of work; we open sourced ts-for-gjs as a by product of the development of this. Still, I see that it's largely not needed; using the GJS constructor with props is a nice way of constructing objects.

I think the option makes sense, as I wrote in my earlier comment, so let's go with that. I was about to merge this in, but I think I notice one small thing that isn't quite right.

Picking a random class; I've chosen Gio.Task. Let's look at the constructors/static functions generated both with and without the "-s" option:

With -s:
```
      static name: string
      static new (config?: Task_ConstructProps): Task
      constructor (config?: Task_ConstructProps)
      static new(source_object: GObject.Object | null, cancellable: Cancellable | null, callback: AsyncReadyCallback | null): Task
      static is_valid(result: AsyncResult, source_object?: GObject.Object | null): boolean
      static report_error(source_object: GObject.Object | null, callback: AsyncReadyCallback | null, source_tag: object, error: GLib.Error): void
```

Without -s:
```
      static name: string
      static new (config?: Task_ConstructProps): Task
      constructor (config?: Task_ConstructProps)
      static is_valid(result: AsyncResult, source_object?: GObject.Object | null): boolean
      static report_error(source_object: GObject.Object | null, callback: AsyncReadyCallback | null, source_tag: object, error: GLib.Error): void
```

And here's what used to happen:
```
  export interface Task_Static {
      name: string
      new (config?: Task_ConstructProps): Task
  }
  export declare class Task_Static {
      new(source_object: GObject.Object | null, cancellable: Cancellable | null, callback: AsyncReadyCallback | null, callback_data: object): Task
      is_valid(result: AsyncResult, source_object: GObject.Object | null): boolean
      report_error(source_object: GObject.Object | null, callback: AsyncReadyCallback | null, callback_data: object, source_tag: object, error: GLib.Error):   void
  }
  export declare var Task: Task_Static
```

You're generating the function `static new (config?: Task_ConstructProps): Task` , which I don't think is correct. Looking with gjs itself:
```
gjs> let Gio = imports.gi.Gio;
gjs> Gio.Task
function Gio_Task() {
    [native code]
}
gjs> Gio.Task.new
function new(source_object, cancellable, callback) {
        /* proxy for native symbol g_task_new(); */
}
gjs> Gio.Task.new({ })
Error: Too few arguments to function Gio.new expected 3 got 1
gjs> let t = new Gio.Task({})
gjs> let t = new Gio.Task()
gjs> Gio.Task.new()
Error: Too few arguments to function Gio.new expected 3 got 0
```

This function was generated before in the interface of the static side as the constructor function. Now that we've moved to using the typescript `constructor` keywork on the class declaration, I don't think we should be generating this anymore.

makepost:
Okay, `-s` is gone, files are big again, and `Gio.Task` now has the following:

```typescript
    static name: string
    constructor (config?: Task_ConstructProps)
    static new(source_object?: GObject.Object | null, cancellable?: Cancellable | null, callback?: AsyncReadyCallback | null):Task
    static is_valid(result: AsyncResult, source_object?: GObject.Object | null): boolean
    static report_error(source_object: GObject.Object | null, callback: AsyncReadyCallback | null, source_tag: object | null, error: GLib.Error): void
```

Reducing size of declaration files is not a priority, just thought they could be made compact. If anyone sees how to work around TypeScript inheriting static methods, would be interesting to see.

sammydre:
Thanks. That works and I've merged it.

Yes, Typescript static method inheritance is the cause of our troubles here. It's a shame the definitions end so large due to the repeated code.

PR #7, opened 2018-11-19
Initialize GirRepository.ns to get rid of warning

realh:
When editing main.ts vim flagged an error about GirRepository.ns possibly not being initialized. That got on my nerves so I got rid of it by adding a minimalist initializer, which seemed to be the most efficient fix.


realh:
This change seems to have accidentally snuck in from one of my other branches.

PR #14, opened 2019-07-11
Signal methods

realh:
[According to the gks wiki](https://gitlab.gnome.org/GNOME/gjs/wikis/Mapping#signals) GObjects have methods for signals, but these were missing from ts-for-gjs' output, so I've added them.

Sorry this can't be merged cleanly. I think it's because I originally added this to the wrong branch, and tried to patch it into a clean branch from upstream afterwards, but something seems to have gone wrong. If it's a problem I'll try doing it over again.


sammydre:
Thanks!

PR #22, opened 2020-08-02
package.json: add ts-node as devDep to fix running in-tree

peat-psuwit:
When oclif detects src/-lib/ split, it'll try to use .ts files in src/
directly. If ts-node isn't installed, it'll fail and won't re-try the
file in lib/, thus failing the run entirely.

Add ts-node as a devDep fixes the issue and allow the command to run
directly in-tree.


JumpLink:
@sammydre looks good to me

PR #23, opened 2020-08-02
template: improve typing for cast.ts

peat-psuwit:
The first improvement is typing `inheritanceTable` as indexable, which
enables us not to cast it to `any` needlessly, improving type safely.

The second improvement is making `StaticNamed` interface a generic and
adding construct signature. This way, TypeScript will be able to infer
the type parameter of `giCast`, which allow users to write:
```js
const buf = giCast(srcView.buffer, GtkSource.Buffer)
```
without writing the type twice.


JumpLink:
Nice, thank you @peat-psuwit 
@sammydre looks good to me, can you merge this?

PR #24, opened 2020-08-02
template: uses <%_ to fix indent of node-gtk's index.d.ts

peat-psuwit:
Without <%_, the preceeding space won't be removed and will be included
in front of any following line. This has significance when outputing
`// <reference` line, as TS (at least in VSCode) won't read the line if
it's indented at all.


JumpLink:
@sammydre looks also good to me

PR #39, opened 2021-01-07
Use /usr/local/share when on Darwin systems

derekstavis:
This enables Darwin systems to work OOB, without needing to specify additional directories on the configuration.


derekstavis:
@sammydre friendly ping 😛

sammydre:
Thanks @derekstavis , a "friendly ping" was just what I needed!

PR #40, opened 2021-02-15
Add support for overloadable methods

JumpLink:
Until now, all method names that already exist in a class are ignored, but they can easily be added because typescript recognises them as overloadable methods.

Other improvements in the examples:
* Upgrade dependencies
* Removed some TODO's which are no longer TODOS
* Replaced google.com with duckduckgo.com (better data protection)
* Ported more examples from node-gtk


JumpLink:
Closed, as I still want to make more adjustments

PR #45, opened 2022-01-17
Moved type definition generation to new TypeDefinitionGenerator

JumpLink:
The parsing and processing of the gir files and the generation of the type definitions should be separated. For this I created a new class "TypeDefinitionGenerator", all methods to create the strings for the type definitions are now included here.

The generated data are now stored in the gir element objects. For example, the following data is now stored in a GirClass:

```ts
export interface TsClass {
    name: string
    qualifiedName: string
    parentName?: string
    qualifiedParentName?: string
    localParentName?: string // qualified if its module != qualifiedName's module
    namespace: string
    version: string
    isDerivedFromGObject?: boolean

    constructPropInterfaceName: string
    inheritConstructPropInterfaceName?: string
    /**
     * `true` if this is a abstract class.
     * E.g GObject.ObjectClass is a abstract class and required by UPowerGlib-1.0, UDisks-2.0 and others
     */
    isAbstract: boolean
    /** Object with all property / method names (Used to prevent duplicate class properties / method names) */
    localNames: LocalNames
    /** Object of all constructor properties (used to generate the constructor property interface) */
    constructPropNames: LocalNames

    /** Fields of the base class itself */
    fields: GirFieldElement[]
    /** Properties of the base class itself */
    properties: GirPropertyElement[]
    /** Constructor properties of the base class itself */
    constructProps: GirPropertyElement[]
    /** Array of property names of the base class itself (used to generate signals) */
    propertyNames: string[]
    /** Methods of the base class itself */
    methods: GirMethodElement[]
    /** Virtual methods of the base class itself */
    virtualMethods: GirVirtualMethodElement[]
    /** Constructor methods of the base class itself */
    constructors: GirConstructorElement[]
    /** Static functions of the base class itself */
    staticFunctions: Array<GirFunctionElement | GirConstructorElement | GirMethodElement>
    /** Signals of the base class itself */
    signals: GirSignalElement[]

    extends: {
        /** module packageName + full system name of extended class */
        [key: string]: {
            /** The extended class itself */
            class: GirClassElement | GirUnionElement | GirInterfaceElement | GirRecordElement
            /** Fields of extended class */
            fields: GirFieldElement[]
            /** Properties of extended class */
            properties: GirPropertyElement[]
            /** Methods of extended class */
            methods: GirMethodElement[]
            /** Virtual methods of extended class */
            virtualMethods: GirVirtualMethodElement[]
            /** Signals of extended class */
            signals: GirSignalElement[]
        }
    }

    implements: {
[...]

PR #47, opened 2022-01-22
[node-gtk] don't add $obj parameter in connect() method either

peat-psuwit:
I used connect() in node-gtk, and there it doesn't pass $obj to the
callback either. Besides, node-gtk's connect wrapper passes the callback
to connect() as-is [1], so the callback signature shouldn't be
different.

This change is also propagated to notify signal methods too, by consolidating
the notify signal method generation with the normal notify signal method generation. This has a small side-effect on node-gtk; see the commit message.

[1]
https://github.com/romgrk/node-gtk/blob/3051382de16bf730716248087137f5cb611d9923/lib/bootstrap.js#L60

### DIFF
--- a/src/template-processor.ts
+++ b/src/template-processor.ts
@@ -97,6 +97,7 @@ export class TemplateProcessor {
         params: string[],
         retType: string,
         identCount = 1,
+        emitable = true,
     ) {
         const ident = this.generateIndent(identCount)
         const def: string[] = []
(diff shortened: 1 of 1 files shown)


JumpLink:
@peat-psuwit Looks good to me, I like that you have removed duplicate code. Unfortunately your PR has some merge conflicts with my PR #45. Would you create a new PR to my fork? Then we can merge both together here. The methods you have changed are located in `type-definition-generator.ts.

peat-psuwit:
I'll take a look and see what I can do. Although I probably won't look into it until the next weekend at the earliest.

JumpLink:
Thanks @peat-psuwit , that would be great 👍

JumpLink:
@peat-psuwit should also be fixed with #45, would you like to see if it works for you now?

Issue #21, opened 2020-04-09
node-gtk typings issues


romgrk:
Hey,

node-gtk maintainer here. I've reviewed some of the typings for node-gtk and here are some issues I've found. I'll refer to `Gtk-3.0.d.ts` as an example. I'll use prefix `g-` when using glib terms and `ts-` when using typescript terms.

## 1. g-interfaces are listed as ts-classes. For example 
```typescript
export class Editable {
     /* Methods of Gtk.Editable */
     copyClipboard(): void
     // ...
}
```
g-interfaces can be implemented by g-classes, so I don't think ts-class is the correct mapping for g-interfaces. For example, classes `Gtk.Entry` and `Gtk.TextView` both implement `Gtk.Editable`.

## 2. g-signals don't have a `connect_after` method
```
     /* Signals of Gtk.Editable */
     connect(sigName: "changed", callback: (($obj: Editable) => void)): number
     connect_after(sigName: "changed", callback: (($obj: Editable) => void)): number
     emit(sigName: "changed"): void
     on(sigName: "changed", callback: (...args: any[]) => void): NodeJS.EventEmitter
     once(sigName: "changed", callback: (...args: any[]) => void): NodeJS.EventEmitter
     off(sigName: "changed", callback: (...args: any[]) => void): NodeJS.EventEmitter
```
`.on` and friends also don't return the instance but I take that as a node-gtk issue as it should respect nodejs semantics, will be fixed soon.

Note however that there is an open PR that will add an optional argument to `.on(signalName: string, cb: Function, after?: boolean)` (and `.once`).
Note that I would also don't tend to promote usage of `GObject#connect`. It's a lower-lever method and all it's use-cases are satisfied by `GObject#on`. Just saying.

## 3. `GObject#on` callbacks arguments are known
```typescript
connect(sigName: "delete-text", callback: (($obj: Editable, startPos: number, endPos: number) => void)): number
on(sigName: "delete-text", callback: (...args: any[]) => void): NodeJS.EventEmitter
```
The arguments for the `.on` callback are the same as for the `.connect` callback.

**Note**: node-gtk skips passing the instance as an argument to the callback because it's a leftover of the C paradigm where closures don't exist. So `.on('delete-text', (editable, startPos, endPos) => { /* ... */ })` becomes `.on('delete-text', (startPos, endPos) => { /* ... */ })`. 

## 5. typings don't seem to use inheritance for g-classes
```typescript
export class AccelLabel {
[...]

Issue #13, opened 2019-07-10
Interfaces


realh:
GObject interfaces being exported as TS classes instead of interfaces can be an inconvenience, requiring one to override type safety with explicit casts.

For usability it would be best if the TS definition referred to a TS interface, ie simply change `class` to `interface`. However, that would hide the JS object we need to use to initialise classes that implement GInterfaces. But they could still be accessed by looking up the names as strings in imports.gi.*. When a new decorator-based API is designed to replace `GObject.registerClass` there's an opportunity to make this a non-issue by making the `g_implements` (or whatever it ends up being called) decorator accept a string as an alternative to a class.

JumpLink:
A decorator based API would be very nice, but I think it will be a long time before ECMAScript decorators are included in the standard.

Conveniently, we are in the TypeScript world where decorators have been used for a long time. So I would like to develop a TypeScript framework where we could implement decorators for GObject classes ourselves. Then it would probably not be a problem anymore if GObject interfaces are also generated as Typescript interfaces.

So the question is, do we want to implement that here or in a separate framework and do we want to continue to support interfaces as classes in ts-for-gir then?

realh:
I've written my own decorators. They would need some work to make them suitable for "production", but they might make a useful starting point.

[gdeco.js](https://github.com/sammydre/ts-for-gir/files/8427508/gdeco.js.txt)
[gdeco.d.ts](https://github.com/sammydre/ts-for-gir/files/8427507/gdeco.d.ts.txt)

I think the gjs developers are open to improving the API for class registration etc in a way that's forward compatible with decorators, so you could try liaising with them.

realh:
On the subject of interfaces as classes, I found that when you define a class in Typescript, it automatically creates an interface with the same name, so you might as well continue to treat them as TS classes, making it easier to support static methods etc. For my inheritance fork, there is effectively no distinction between GObject classes and interfaces. Each one is declared both as a TS interface containing the instance methods and properties, and a TS object (not a class) containing the static methods etc. This almost completely solves the problem with clashing overloads, but there are a few cases that need workarounds.

JumpLink:
Thanks a lot, I think this information will help me a lot to be able to implement the whole thing, I will also take a closer look at your fork as soon as I tackle this.

Submitting the decorators directly to Gjs is a good idea, but I will have to experiment with it myself beforehand to provide a clean implementation.

[...]

Issue #12, opened 2019-07-10
No class hierarchy


realh:
I noticed that all GObject-based classes are exported without parent classes. Instead, all members etc of the parent GObject-class are copied into the definition of the new class. This causes problems when you want to pass one of these classes to a function that expects a parent class.

Is there an important reason you made it do this instead of using extends?

nykula:
Look at #3, we tried to replicate the hierarchy to stop ending up with a gazillion megabytes of typings to parse on every check or IDE launch, but had trouble with static members. TypeScript inheritance and GObject inheritance aren't compatible when it comes to that. Maybe you'll find a fix, please see.

realh:
I've read the comments for #3, but I still don't understand why static method inheritance is a problem. Could you explain that more?

nykula:
When a GObject subclass had a static method with same name and argument count as one in its parent classes, but their argument or return types differed, we got either typings that didn't compile or code that passed type checks but didn't work in Gjs. Unfortunately I don't remember more specifics.

realh:
OK, I see that Typescript doesn't allow even static members to override parents with incompatible types. I'm not very familiar with the language, but I think I've found a workable solution. I haven't tested it yet with any real-world classes from GObject, but a simple test case compiled OK in tsc, and it generated code that did what I expected.

Say what we want is this:

    export class Parent {
        myMethod(): void
        static new(blurb: string): Parent
    }

    export class Child extends Parent {
        myMethod(): void
        static new(blurb1: number, blurb2: number): Child
    }

The compiler rejects that, but accepts it if I simply add this to `Child`:

        static new<T, V>(blurb1: T): V

Surprisingly, I don't need to change `Parent`, which makes things considerably simpler, nor do I have to provide a second argument in the generic method signature. I have to admit I don't know why it's lax about that, but it's good for us. I should probably do some more research on stackoverflow, and also work out what happens if `Parent.new` takes more arguments than `Child.new` instead of the other way round.

Using a generic for the return type actually tightens up type safety, because if you try to abuse it something like this:

    const child = Child.new("blurb")

child would be of type `unknown`, most likely causing errors when we try to use it. The error would be in the wrong place, but that's better than the compiler not detecting the mistake at all.

realh:
[...]

PR #63, opened 2022-04-08
Real inheritance and implementations

JumpLink:
Now that the code structure is much clearer, it's time to work on inheritance and implementations :)

Thereby I will be inspired by the ideas of @realh's fork and maybe adopt them.

Related tickets
* #13
* #12
* #21 (Point 5)


JumpLink:
@realh I am excited about how well this seems to be working with your trick to just use an interface with the same name of the class to allow multiple inheritance. This seems to work also with classes, so it is not necessary to define a property for it.

I think I also have a good idea how we can solve the problem of incompatible properties.

I hope we can soon work on the same code base again.

PR #66, opened 2022-05-22
Refactor Ava test to work with the new TS generation logic

HeavenVolkoff:
Hello,
 
This PR attempts to address the pending TODO, in [NEWS](../blob/master/NEWS.md), to refactor the Ava test. I decided to solve this to better accommodate myself with the code base and understand how this project works (Great work btw).
I had to update the testData.ts by hand to deal with the deprecation of `allow-none`. However, I am not sure if the changes I made are 100% correct, as this is my first time dealing with GLib, GIR and Gjs. Besides that, there are a lot of changes in this file, but they are all consequence of prettier formatting, if these are too cumbersome I can revert them. 
Also, I changed some test cases results to accommodate some minor changes in how the TS code is generated now. Mostly, it was fixing the comments, adding namespace prefix and dealing with the changes in how `construct-only` is currently dealt with (#52).
Beside these changes, I also updated the vala-girs submodule, if this is undesired I can also revert it.


JumpLink:
Thank you very much! I need to look at this more in detail. I was also thinking about using something completely different than ava for the tests. After I'm done with my current PR #63 the test will probably need to be adjusted again anyway, so I'll wait with the review yet

JumpLink:
This would be great, you do not need to close it. Thank you!

JumpLink:
@HeavenVolkoff Okay now you can continue :)

I am very happy that you want to contribute to this project, if you have any questions or need tasks, feel free to let me know. :+1:

HeavenVolkoff:
Rebased the PR on the current master.

Most tests are failing, but thankfully no errors, just output differences, which is expected now that the project is using inheritance.

The only weird thing is the `function` and `callback` tests, which are getting empty data from the generator.

But overall, the rebase was mostly painless.

JumpLink:
Like that's great, I can take another look at that if you like. And as soon as everything works out we can merge this.

But I have thought about the ava test again. I see the following disadvantages of the ava test:

* The order of the methods and properties should not matter, it should only matter that they are there and correct.
* It is tedious to add new tests because the gir XML is hard to write by hand.
* When errors occur it tempts the developer to just copy the current result to make it fit, but this is not the purpose of such tests.

Therefore I would suggest to rethink the test, I would have the following idea:

* Create simple vala classes for each test (see https://github.com/antono/vala-object to see how this can be implemented)
* build the gir files from that
* Run the generation logic / parser on that
* Test if the desired data was parsed correctly

This way we could for example write Vala classes with methods that have certain parameter types and test if the parameters have the correct type in Typescript and if all methods are present.

The type generator could be tested separately by simply passing it the necessary data and seeing if the correct typescript is returned


What do you think?

PR #67, opened 2022-05-22
Add typing definition for `imports.package`

HeavenVolkoff:
Hello,

This PR aims to implement a type definition for [imports.package](https://gjs.guide/guides/gtk/application-packaging.html)

I added the `Package` namespace in Gjs.d.ts and exported it in Gjs.js, also added a definition of the `package` type in the  global `imports` variable.
Beside these changes, I also fixed what seems to be a typo in the export definition for `_imports` in index.d.ts, and ran prettier in the files to fix some style inconsistencies.

PR #68, opened 2022-05-26
Revert removal of imports.versions

HeavenVolkoff:
Hello,

This PR intends to revert a [mistake I made](https://github.com/sammydre/ts-for-gir/pull/67/files#diff-a6b8083e7461cce7cc3413ca471e7d9d0a8bdf18911a2a6e777c0fcbc689f12bL260-L269) in #67, which was the removal of the global definition for `imports.verision`


JumpLink:
Thank you, I had overlooked that

PR #69, opened 2022-05-26
Tell tsc to expect error on gi ESM imports

HeavenVolkoff:
Hello,

Currently, when using the type definitions generated by this tool with the Typescript compiler it complains about not being able to find the modules imported by the generated .js files

image

This is an expected behaviour considering TS doesn't understand `gi://` imports, and that doesn't really impact its type checking due to it resolving everything through the `d.ts` files.

So, this PR adds to the `module.js` template, used to generate the `.js` files, a `@ts-expect-error` annotation on top of the `gi://` imports to avoid TS compiling errors


JumpLink:
Thank you

PR #70, opened 2022-06-30
Add GLib.Variant unpacking methods

HeavenVolkoff:
Hello,

This PR adds the definitions for the unpack methods available to GLib.Variant instances.

Their behaviour is described here:
https://gjs.guide/guides/glib/gvariant.html#unpacking-variants

They are not present in the gir, due to being gjs specific:
https://gitlab.gnome.org/GNOME/gjs/-/blob/1.72.0/modules/core/overrides/GLib.js#L301-313


JumpLink (changes requested):
@HeavenVolkoff good work, thank you!

I have create a [PR](https://github.com/HeavenVolkoff/ts-for-gir/pull/1) for your PR as a change request, you can just merge them, but this way you can better take a look at it ;)

HeavenVolkoff (commented):
Merged you PR

JumpLink:
> Merged you PR. Thanks for reviewing and improving the code. Next time I'll try to implement examples for new type definitions.

Perfect, thank you!

PR #83, opened 2022-09-13
Follow user defined verbose option in ConflictResolver

HeavenVolkoff:
Hello,

This PR changes ConflictResolver internal logger to follow the user defined verbose configuration.


JumpLink:
Good catch, thank you :+1:

PR #84, opened 2022-09-13
Fix prepend logic in Logger's static methods

HeavenVolkoff:
Hello,

This PR fix a bug in Logger's static methods that result in the messages being printed twice:
image
With the following changes:
image


JumpLink:
Thank you :)

PR #85, opened 2022-09-13
Expose GIRepositoryNamespace underscore properties in TS namespace

HeavenVolkoff:
Hello,

In GJS's source code, all GIRepositoryNamespace (Object representing imported GIR libraries) expose two underscored properties containing the lib name and version:
https://gitlab.gnome.org/GNOME/gjs/-/blob/master/gi/ns.cpp#L186-190

This PR exposes those properties in the Typescript namespace definition for each processed lib.


HeavenVolkoff:
Will check why the tests are failing

HeavenVolkoff:
Fixed the tests.

This was originally a very quick and dirty change to add the support needed for a project I am working, so I only tested on the environment that I was developing in, sorry. It should now be a cleaner implementation with support for all the environment this tool is targeting.

Issue #72, opened 2022-06-30
Type conflicts with lib.dom.d.ts


HeavenVolkoff:
Hello,

Currently, the type definition for the global `console` is commented out in:
https://github.com/sammydre/ts-for-gir/blob/979d2a016f182ca54f14047483d2b29cac126ade/templates/Gjs/index.d.ts#L201-L202

This is due to conflicts with the global definitions shipped with typescript, that are used by most IDEs (such as VSCode):
https://github.com/microsoft/TypeScript/blob/93f2d2b9a1b2f8861b49d76bb5e58f6e9f2b56ee/lib/lib.dom.d.ts#L17750

However, this results in gjs code being linted with `lib.dom.d.ts` type definitions, which is not ideal, as gjs implementation of those interfaces has slight differences and incompatibilities with the DOM implementation present in browsers, and the vast majority of the DOM API is not implemented in gjs (and probably will not be, according to [#265](https://gitlab.gnome.org/GNOME/gjs/-/issues/265)).

This can be worked around by using a custom tsconfig.json / jsconfig.json that doesn't contain DOM in the lib property:
```
{
  "compilerOptions": {
    "lib": ["esnext"],
    "allowJs": true,
    "checkJs": true
  },
  "include": ["src/**/*"],
  "exclude": ["node_modules"],
  "$schema": "https://json.schemastore.org/tsconfig"
}
```

So far in my tests with VSCode, this prevents the `lib.dom.d.ts` from being used, and allows the local definitions generated by ts-for-gir to work without error, even with `console` commented out and my, yet not upstreamed, [additions of typing definition for gjs recently implemented Text(D)Encoder ](https://github.com/sammydre/ts-for-gir/compare/master...HeavenVolkoff:ts-for-gir:add-text-encoder-decoder).

IMHO, ts-for-gir could attempt to solve this automatically. The logic I am considering would be something along the lines of:
- If the project where the definitions are being generated doesn't have a `tsconfig.json` than generate one following the example above
- If the project already has a `tsconfig.json`, check if DOM is included in the `lib` property:
  - If it is not, then do nothing and continue
  - If it is, either:
    - Throw an error and stop
    - Throw a warning and keep current behaviour (with conflicting globals not being included in the generated types)

JumpLink:
Hello @HeavenVolkoff,

[...]

PR #86, opened 2022-09-13
Solution for #72 + Add type definitions for Text(D)Encoder

HeavenVolkoff:
Hello,

This PR presents a solution for #72. It adds a notice to the README informing users that it is recommended to edit their `tsconfig.json` file for GJS projects. Besides, logic was built to check, during runtime, if the project has the `DOM` lib in its `tsconfig.json` and if that is the case ask the user if he wants to skip generating any conflicting type.

Also, included in this PR are the definitions for the Text(D)Encoder interfaces added in GJS 1.69.2:
https://gitlab.gnome.org/GNOME/gjs/-/blob/572f8e8c/NEWS#L384-393
https://gitlab.gnome.org/GNOME/gjs/-/blob/master/doc/Encoding.md
https://gitlab.gnome.org/GNOME/gjs/-/blob/master/modules/esm/_encoding/encoding.js

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,11 @@ And then create a new script in your `package.json` to generate the types, for e
 
 more examples can be found in the [./examples/](/examples/) subfolder.
 
+### For GJS projects
+It's recommended that you create or modify your `tsconfig.json`/`jsconfig.json`, so it doesn't include the `DOM` lib, as it conflicts with some generated GJS global types and will cause lint warnings and compilation errors with typescript.
+
+Either add/edit the `lib` property so it doesn't include `"DOM"`, or enable the property `noLib` (However the side effects of doing this should be considered). For more information check the documentation for both the [`lib`](https://www.typescriptlang.org/tsconfig/#lib) and [`noLib`](https://www.typescriptlang.org/tsconfig#noLib) properties.
+
 ### CLI
 
 To generate the Typescript type definitions of Gtk-4.0 for Gjs run
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -52,8 +52,7 @@
     "eslint": "^8.20.0",
     "eslint-config-prettier": "^8.5.0",
     "eslint-plugin-prettier": "^4.2.1",
-    "rimraf": "^3.0.2",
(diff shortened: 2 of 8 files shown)


HeavenVolkoff:
Will check why the tests are failing

HeavenVolkoff:
So, typescript has its own parser for `tsconfig.json`, and it seems to accept more than the pure JSON spec, because of this using `JSON.parse` is not possible as it can fail with valid `tsconfig.json`. My solution was to simply import the typescript compiler's own parse and use it directly. The consequence is that this requires typescript to be a runtime dependency, IMHO I don't think this would have much of an impact for end users, as this tools will probably be a `devDependency` on project using it and most of them will probably also use typescript. However, if you don't agree with this, I can look for alternative parser that may offer a better compatibility with  `tsconfig.json`, [json5](https://www.npmjs.com/package/json5) seems to be a good alternative.

Also, I don't know what happened with the tests in the latest commits, running them on my machine resulted in no error. Could you maybe rerun the actions for those that failed?

HeavenVolkoff:
> @HeavenVolkoff Another alternative is [jsonc-parser](https://www.npmjs.com/package/jsonc-parser) which is from Microsoft, so maybe this package is also used internally in TypeScript?

Although [jsonc-parser](https://www.npmjs.com/package/jsonc-parser) seems to be a good option, it isn't what Typescript uses internally, they actually implemented their own parser, because they seem to try parsing some broken JSON without failing. Here you can see their parser implementation:
https://github.com/microsoft/TypeScript/blob/main/src/compiler/parser.ts

> **Edit** I just realized that the tsconfig.json can also inherit from other config files using `extend`, to parse this should work 
better with your current solution so I will merge that now :)

Oh, that is a good point, I didn't consider it when I implemented this XD. Just went with the simplest solution. But, I would say it is the main advantage of using TS own parsing logic, even if they change format or break compatibility, it would have very little effect on this project.

> P.S. Today is my birthday and I just saw your pull requests today, which is a great birthday present, thank you! :)

Hey man, Happy Birthday 🎂. Hope you enjoy your day.

Issue #87, opened 2022-09-17
`Gio-2.0.Cancellable.connect` is typed incorrectly for GJS


HeavenVolkoff:
Hello,

Currently, the generated definitions for `Gio-2.0` on GJS has incorrect typing for the `Cancellable.connect` method according to GJS's [documentation](https://gjs-docs.gnome.org/gio20~2.66p/gio.cancellable#method-connect) and directly testing on its REPL:
![image](https://user-images.githubusercontent.com/9082460/190839443-d3563530-22a4-4e19-a5ef-fee74476b160.png)

It seems `Gio-2.0.Cancellable` doesn't follow the `GObject.connect` convention and implements a custom `connect` as far as I could understand.

To solve this I tried to simply [implement an injection into  `Gio-2.0.Cancellable` to override the `connect` method](https://github.com/sammydre/ts-for-gir/commit/4ffc6d8be63f69bf8da5c48cebfbd8347a5ce97f), however that resulted in unresolved conflicts and ultimately all implementations of `Gio-2.0.Cancellable.connect` ended up commented out of the type definitiondue to conflicts:
![image](https://user-images.githubusercontent.com/9082460/190839556-3693c767-7585-42e1-8b0e-9f358ca99158.png)

I then tried to understand why this was happening in [`conflict-resolver.ts`](/sammydre/ts-for-gir/blob/master/packages/cli/src/conflict-resolver.ts), and from what I could gather this seems to be a temporary solution:
 https://github.com/sammydre/ts-for-gir/blob/bdfd98a5654a8ffcaafe623ae123981ad39429d1/packages/cli/src/type-definition-generator.ts#L545-L548
 
I don't know if the injection solution and solving the current method conflict shortcoming is the correct path for this particular problem, or if there is some other, simpler, way. Maybe this could be solved by adding an exception on commenting out conflicting _injected_ methods? Any ideas?

HeavenVolkoff:
I was able to make the injection solution work by adding the exception on commenting out conflicting injected methods, and using an override to the connect method, it doesn't exactly match the runtime behaviour, but at least it does not complain on correct code anymore:
https://github.com/sammydre/ts-for-gir/compare/master...HeavenVolkoff:ts-for-gir:fix-gio-cancellable-connect

[...]

PR #90, opened 2022-09-24
Fix `Gio.Cancellable.connect` method signature

HeavenVolkoff:
Workaround for #87 

`Gio.Cancellable` extends  `GObject.Object`. However, `Gio.Cancellable.connect` method doesn't follow the `GObject.Object.connect` signature: 
https://gjs-docs.gnome.org/gio20~2.66p/gio.cancellable#method-connect

This PR implements the correct signature for the `connect` method as an overload in `Gio.Cancellable` class. This is not completely correct as it still considers the `GObject.Object.connect` signature valid, even tough it throws in runtime. However, it was the only non hacky solution that I could come up with due to how strict Typescript is with inheritance rules.

Beside this change, this PR also implements an exception during conflict resolution for injected methods and improve some of the injected types to be more lax and accept `InjectionFunction` in more places


HeavenVolkoff:
Ignore this. Check #91 for a better approach.

PR #91, opened 2022-09-25
Change Class Signal conflict resolution logic

HeavenVolkoff:
Hello,

This is a better solution for #87.

It changes the conflict resolution logic for classes signals. Now it accepts methods as overloads for parent signals, and pull incompatible parent methods as overloads for signals. As far as I could test, this allows for classes that implement `methods` with incompatible signature to `propertySignalMethods` be properly typed. However, with the caveat that classes that completely override the parent implementation (such as `Gio.Cancellable.connect`) will still be typed as accepting it, even tough that is not true on runtime. I am thinking on a solution for this, but it will probably require some runtime monkey-patching, so I don't know if it will be viable. For now, I think this is a good approach to solve the original issue tough.

### DIFF
--- a/packages/cli/src/conflict-resolver.ts
+++ b/packages/cli/src/conflict-resolver.ts
@@ -1007,7 +1007,7 @@ export class ConflictResolver {
                     // Function vs. Signal
                     else if (this.tsElementIsSignal(b.data)) {
                         this.log.debug(`${className}.${name} Direct Function vs. Signal`)
-                        baseFunc.hasUnresolvedConflict = true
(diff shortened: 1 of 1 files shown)


JumpLink:
I have tested your branch and that looks good to me, thank you again :)

The only thing that is a bit stupid is that the method is now overloaded for each signal, while once would be enough.

See screenshot:
grafik

We should make an issue for that.

Your idea to disallow the signatures (using ESLint) that are necessary for the types but do not exist is also a very good idea.

Issue #92, opened 2022-09-26
Redund overloads from signal `connect` method


JumpLink:
This happens since #91

![grafik](https://user-images.githubusercontent.com/1073989/192230697-4843484b-155a-4c5e-ab88-a47c9d3d65d4.png)

HeavenVolkoff:
This issue completely slipped though my tests, thanks for finding it. I already came up with a solution that seems to be working fine, but this time I am going through the diffs of all definitions generated by `yarn run build:types:all` to ensure nothing odd comes up. Only downside is that generation got slightly slower, but not by much:

Before:
![IMG_20220927_041823_416.jpg](https://user-images.githubusercontent.com/9082460/192459831-5b8d18be-4759-42bb-b952-72f06a9c14e5.jpg)

After:
![IMG_20220927_041826_311.jpg](https://user-images.githubusercontent.com/9082460/192459881-6918e6bb-1d60-4529-a483-25bd247de32c.jpg)

Also, while checking the diffs, some other cases of repeatead method definitions came up. It seems the node-gtk conflict resolution for methods was adding clashing definitions as overloads even when they were the same. I am going to fix that too. There were some other cases, but I couldn't pinpoint the origin for their occurrence, however my current solution will take care of them as well.

PR #95, opened 2022-09-29
Implement logic to check previously generated methods to avoid repetition

HeavenVolkoff:
Hello,

This is my implementation of a fix for #92, It also includes some other changes:
 - Move functionHasConflict and related functions to utils.ts
 - Break conflict resolver outer loop as soon as unresolved conflicts are detected
 - Improve resolution between class methods and conflicting inherited ones. Only add as overloads inherited methods that are incompatible with the class implementation
 - Add ignore property to TsFunction, differ from hasUnresolvedConflict as no reference to this function will be included in the generated type definition
 - Introduce tracking of generated functions in type-definition-generator.ts's generateFunction to avoid producing redundant entries

I am opening this as a draft for now, because I think this still needs more tests, and I wanted @JumpLink's input about whether he agrees with this solution, as it has a (minor) performance impact.

Another thing I wanted to discuss is if the case of overlapping input/return types in overloads should be dealt by these changes, or if it should be a separated issue (IMO I think so). This seems to be somewhat common, especially with optional and nullable types and sometimes with overlapping union types
Example (notice the optional for the cancellable argument is the only difference between both `init` definitions):
image

### DIFF
--- a/packages/cli/src/conflict-resolver.ts
+++ b/packages/cli/src/conflict-resolver.ts
@@ -1,7 +1,15 @@
 import { GirFactory } from './gir-factory.js'
 import { Logger } from './logger.js'
 import { NO_TSDATA } from './messages.js'
-import { isEqual, merge, clone, typeIsOptional } from './utils.js'
+import {
+    isEqual,
+    merge,
+    clone,
+    typeIsOptional,
+    functionHasConflict,
+    typesHasConflict,
+    paramsHasConflict,
+} from './utils.js'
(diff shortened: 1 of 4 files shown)


JumpLink:
@HeavenVolkoff this all sounds very good so far, unfortunately I don't have time to look at it in detail right now, I hope I'll get to it next week

HeavenVolkoff:
No worries @JumpLink. I am in no hurry. Btw, thank you for the work you put into this project, and in keeping up with my stream of issues and PRs 👍🏼

PR #96, opened 2022-10-26
Replace typescript runtime dependency with get-tsconfig

HeavenVolkoff:
Small improvement over #86.

Replaces the runtime dependency on [typescript](https://github.com/microsoft/TypeScript) with [get-tsconfig](https://github.com/privatenumber/get-tsconfig), a compliant tsconfig.json's parser without the weight of the full typescript compiler. Besides, updates some of the dependencies and add two development dependencies used by this project's root eslint configuration that were missing: `@typescript-eslint/eslint-plugin`, `@typescript-eslint/parser`

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,13 +94,15 @@
   "homepage": "https://github.com/sammydre/ts-for-gjs#readme",
   "devDependencies": {
     "@ts-for-gir/cli": "workspace:^",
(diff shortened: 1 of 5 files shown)


JumpLink:
Nice, thank you! I hope I will have more time soon, then I will have a closer look at the rest of your PR's :)

PR #104, opened 2023-02-01
Add promisified function generation

CharlieQLe:
This adds the function definitions for async/finish functions that can be promisified via `Gio._promisify`. This does not mean that the functions are already promisified however, only providing the definition in the event someone does promisify the functions, so it should be made clear that `Gio._promisify` still needs to be called, though we could possibly generate a typescript file that imports the modules and has a function to promisify everything. While I believe this PR already handles most, if not all, of the intended promisified targets properly, I am unsure if this is the implementation that should be shipped with the package.

Reference: https://gjs.guide/guides/gjs/asynchronous-programming.html#promisify-helper


JumpLink (commented):
Thanks for the further PR, I'll take a closer look at it, I just noticed that we can write a good TSDoc comment for it

JumpLink:
>  though we could possibly generate a typescript file that imports the modules and has a function to promisify everything.

Not a bad idea 🤔, I think we can do it.

JumpLink (changes requested):

examples/Gjs/gio-2-cat-promisify/main.ts
```diff
@@ -0,0 +1,34 @@
...
+import Gio from "gi://Gio";
+
+const loop = GLib.MainLoop.new(null, false);
+
+Gio._promisify(Gio.File.prototype, 'load_contents_async', 'load_contents_finish');
+
+async function cat(filename: string) {
+    const file = Gio.file_new_for_path(filename);
+
+    // TODO: fix Promise return type
```
JumpLink, line 23: Here the wrong return value is currently still returned

JumpLink (changes requested):

JumpLink:
Thank you very much! I've added another example to test your PR with. Unfortunately it seems that the wrong return value is returned here

packages/cli/src/type-definition-generator.ts
```diff
@@ -590,6 +591,82 @@ export default class TypeDefinitionGenerator implements Generator {
     ) {
         const def: string[] = []
 
+        if (this.config.promisify) {
```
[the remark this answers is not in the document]
JumpLink: Okay I have done the work for this

examples/Gjs/gio-2-cat-promisify/main.ts
JumpLink: Should also be fixed now

JumpLink:
Is merged now! 🚀

> though we could possibly generate a typescript file that imports the modules and has a function to promisify everything.

We can do / discuss this in a separate PR 👍

PR #106, opened 2023-02-15
Add support to generate a package for each type we can publish on NPM

JumpLink:
- [x] Add a new CLI option to generate NPM packages
- [x] Add a new CLI option to change the NPM package scope name
- [x] Documentation for the new CLI options
- [x] Generate a package.json for each package
- [x] One folder for each package
- [x] Add an example
- [x] Add dependencies to each package.json
- [x] Package names in lower case
- [x] Separate package names for node-gtk types
- [X] Package directory and file names in lower case
- [x] Parse library versions
- [x] Add a README.md for each NPM package
- [x] Different README.md for node-gtk NPM packages
- [x] Add an test to CI
- [x] Fix tsconfig.alias.json
- [x] Add support for [ambient modules](https://www.typescriptlang.org/docs/handbook/modules.html#ambient-modules)
- [x] Documentation for ambient modules
- [x] Generate both CJS and ESM modules
- [x] Add node-gtk to the peer dependencies to the node-gtk NPM packages 
- [x] Remove destination environment directory
- [x] Separate README.md for the main NPM packages (`@gir/gjs` and `@gir/node-gtk`)
- [x] Generate and publish the new [NPM packages](https://www.npmjs.com/search?q=%40girs)
- [x] #114
- [x] Test the published packages
- [x] Include only desired types (instead of all existing ones). For this, a global interface named GjsImports should be used (see also https://github.com/gjsify/gnome-shell/issues/1)
- [x] Add repository and homepage to package.json
- [x] Documentation for the new NPM packages
- [x] Move the GNOME Shell type definitions to a [NPM package](https://github.com/gjsify/gnome-shell) and make use of the generated NPM packages
- [x] Update NEWS.md
- [ ] ~Auto generate promisified functions?~ (#121)


CharlieQLe:
Testing out the packages in a project I have and it seems like `@girs/xlib-2.0` is not a package, but `@girs/node-xlib-2.0` is.

JumpLink:
Thanks! I will probably not continue for a few days but then I look at it this

JumpLink:
@CharlieQLe I have published many missing packages, the release process was likely terminated accidentally.

CharlieQLe:
Packages seem to be good, from my limited testing at least. This might be out of scope since it requires the usage of another function, but including promisified functions would be nice, just to avoid regenerating the types.

JumpLink:
That's actually something I've thought about as well.

PR #111, opened 2023-03-20
Allow Template to be a string

CharlieQLe:
Template needs to be a string for using paths to UI files instead of loading the data in separately.


JumpLink:
Thank you 👍

PR #113, opened 2023-03-31
Improve promisification by detecting types

CharlieQLe:
Async functions that can be promisified do not always end in `_async/_begin` and `_finish`. An example of this is the new `Gtk.FileDialog` class, with the function `open` being asynchronous (Even with this PR, the function still isn't being generated correctly, but given that the docs have the same problem, this is likely not limited to this project).

Instead, check to see if the `async` function ends with a `Gio.AsyncReadyCallback` and if the `finish` function has a return type as well as exactly one parameter of type `AsyncResult`.


JumpLink:
@CharlieQLe Thank you, a [test seem to fail](https://github.com/gjsify/ts-for-gir/actions/runs/4576799952/jobs/8081408647?pr=113), looks like signals are converted to promises here?

For example:

```
Error: @types/Gjs/Camel-1.2.d.ts(2656,11): error TS2430: Interface 'NetworkService' incorrectly extends interface 'Service'.
  Types of property 'connect' are incompatible.
    Type '{ (sigName: "notify::connectable", callback: ($obj: NetworkService, pspec: ParamSpec) => void): number; (io_priority: number, cancellable: Cancellable | null, callback: AsyncReadyCallback<...> | null): void; (sigName: "notify::host-reachable", callback: ($obj: NetworkService, pspec: ParamSpec) => void): number; (io_...' is not assignable to type '{ (io_priority: number, cancellable: Cancellable | null, callback: AsyncReadyCallback<this> | null): void; (io_priority: number, cancellable: Cancellable | null): Promise<...>; (sigName: "notify::connection-status", callback: ($obj: Service, pspec: ParamSpec) => void): number; (sigName: "notify::display-name", callb...'.
      Types of parameters 'sigName' and 'io_priority' are incompatible.
        Type 'number' is not assignable to type '"notify::connectable"'.
```

Can you check what's going on here? I guess here methods are promisified that cannot be promisified. Can you check this out? As soon as I have time I can take a closer look myself, but right now I don't have that time.

CharlieQLe:
Seems like some of the modules have an async connect function that unfortunately matches the name of the existing GObject connect function. The only solution I can really think of is to avoid generated promisified connect functions. This is probably fine, since the GJS docs do not mention these async functions.

JumpLink:
@CharlieQLe Thank you, can you also check if `contains_email` and `add_timezone` should actually be overloaded with the promisified version? If so, I can see if I can fix that conflict

CharlieQLe:
> @CharlieQLe Thank you, can you also check if `contains_email` and `add_timezone` should actually be overloaded with the promisified version? If so, I can see if I can fix that conflict

It should, but BookBackendSync and the other Sync interfaces actually uses the same function name as the async `contains_email` and `add_timezone` for their synchronous functions. `BookBackend` does provide a synchronous function for `contains_email` though (same goes for `CalBackend`), so perhaps we could avoid generating `BookBackendSync` and the other offending interfaces? If we do go down that route, we should supply the generator with a list of blacklisted modules, interfaces, classes, and functions in order to make this more generic for any GIR files that would need it. I don't use these libraries though, so I don't entirely know if the Sync interfaces provide any extra functionality that the non-Sync ones don't already provide.

nahuelwexd:
I think this is relevant here :eyes: https://gitlab.gnome.org/GNOME/gobject-introspection/-/issues/28

JumpLink:
Vala has also some heuristics to discover begin/finish functions. but a lot is manually downstream annotated through metadata: https://www.vala-project.org/doc/vala/GIDL_metadata_format.html

For example, there are the following relevant properties in the metadata:
* `async` - Force async function, even if it doesn't end in _async
* `finish_name` - Sets custom asynchronous finish function.

The metadata files of Vala can be found here: https://gitlab.gnome.org/GNOME/vala/-/tree/main/vapi/metadata

We may be able to parse this metadata when parsing the gir files as well.

Issue #125, opened 2023-06-02
Dependencies: Can't upgrade xml2js


JumpLink:
See https://github.com/Leonidas-from-XIV/node-xml2js/issues/682

PR #128, opened 2023-06-16
Add generator based on @gi.ts/lib

ewlsh:
Initial generator based on alpha release on @gi.ts (with various changes to integrate with ts-for-gir)


ewlsh:
Experimenting with using `@gi.ts/parser` instead of `xml2js`. It uses https://www.npmjs.com/package/fast-xml-parser instead of xml2js because I recall running into issues with xml2js and some GIR files :thinking: 

Potentially related: https://github.com/gjsify/ts-for-gir/issues/125

JumpLink:
Hey @ewish,

You're mega! I think we can totally ditch the support for node-gtk and instead create a branch called 3.x with the old generator specifically for our node-gtk users. That way, we can seamlessly integrate everything and not worry about juggling both generators.

I'll jump in and lend a hand as soon as possible. Originally, I just wanted to whip up the TypeDoc-generated API docs for the new packages and bring the dependency up to date. But darn it, it hit a roadblock because TypeDoc went through some serious changes.

I am very happy that we are leading the projects together and I will join in as soon as possible :)

With xml2js 0.5.0 has also already had breaking changes, probably it is because of that. But if fast-xml-parser works well, it doesn't matter.

JumpLink:
@ewlsh Do you know why the GitHub action is not running in this branch? You would need to have the appropriate permissions to look at it. I have not found it

ewlsh:
> @ewlsh Do you know why the GitHub action is not running in this branch? You would need to have the appropriate permissions to look at it. I have not found it

I don't know, that is odd :thinking:

ewlsh:
@JumpLink I think there was an issue with the new unified file, I added triggers for PRs and it seems to be running again

PR #134, opened 2023-09-08
fix(node-gtk): override array type only on gjs, fixes #127

### LINKED ISSUE #127: node-gtk: GLib.Bytes.getData() (and perhaps others) returns a JS array, not a Uint8Array
node-gtk doesn't seem to have a special case for guint8[] -> Uint8Array.

```
$ node
Welcome to Node.js v18.16.0.
Type ".help" for more information.
> const gi = require('node-gtk')
> const GLib = gi.require('GLib', '2.0')
> let b = new GLib.Bytes("AA")
> b.getData()
[ 65, 65 ]
> Array.isArray(b.getData())
true
```

I'm not sure if this is a good idea for node-gtk, but since it's how it is in node-gtk, ts-for-gir should produce a correct type.

peat-psuwit:
Node-gtk doesn't do array type override.

### DIFF
--- a/packages/lib/src/gir-module.ts
+++ b/packages/lib/src/gir-module.ts
@@ -598,7 +598,7 @@ export class GirModule {
             arrayType = typeArray[0]
         }
 
-        if (isArray && arrayType?.$?.name && ARRAY_TYPE_MAP[arrayType.$.name]) {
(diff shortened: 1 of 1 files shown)


JumpLink:
Thank you!

PR #144, opened 2024-02-10
gi.ts rework & merge

ewlsh:
This is the second attempt at merging the two codebases, it is still a draft but it is generating types with only a few failures to debug :partying_face:


ewlsh:
I'm working on merging the `newlib/` folder into the rest of the source.

I used `newlib/` for the gi.ts sources so it was easier to merge the codebases side-by-side.

JumpLink:
Yay I'll have a look when I find the time. Shall we delete branch 4.x for this and create a new one?

ewlsh:
> Yay I'll have a look when I find the time. Shall we delete branch 4.x for this and create a new one?

@JumpLink I think it'd be best to create a new branch off of `3.x` called `4.x` or `next` and we can target that branch with this MR.

JumpLink:
@ewlsh I have now recreated the `main` branch for this because I have already referred to it in the documentation. I could start helping by removing the code for `node-gtk`, `CommonJS` and other exports as NPM package and also adjust the examples accordingly. How do we want to do this? Should I create a PR for `gi-ts-rework` or should I push directly to it, which do you prefer?

JumpLink:
#148 solves the current merge conflicts

PR #162, opened 2024-04-11
copy command

JumpLink:
This PR adds a new `copy` command to scan for *.gir files and copy them to a new directory. We want to use this to copy the newest GIR files found on the developer machines to this repository.

To copy all the latest GIR files from your machine into this repository just run `yarn copy:girs` .

## Changes
* Moved XML parsing to DependencyManager to be able to compare the library versions of the found dependencies
* Allow glob patterns for GIR directories to search in
* Allow glob patterns for GIR ignore list
* New `copy` command
* General clean-up like removing unused code
* Split module loading and parsing to omitted the parsing for the `list` and `copy` commands
* Logger output improved and more used 
* Compare GIR library versions and use the latest one
* Copied the GIR files of my machine into this repository
* Bump version `4.0.0-beta.3` 
* Published `ts-for-gir v4.0.0-beta.3` of this PR with the `next` tag
* Published all generated types of this PR with the `next` tag
* Added a helper shell script to find all Fedora packages which contains at least one *.gir file


JumpLink:
@ewlsh Should be finished now, feel free to also update the GIR files from you machine 👍

JumpLink:
@ewlsh ping

JumpLink:
Positive feedback for the packages published of this PR on https://github.com/gjsify/gnome-shell/issues/39 👍

.ts-for-gir.copy-all.rc.js
```diff
@@ -0,0 +1,17 @@
...
+        '/usr/local/share/gir-1.0',
+        '/usr/share/gir-1.0',
+        '/usr/share/*/gir-1.0',
+        '/usr/share/gnome-shell',
+        '/usr/share/gnome-shell/gir-1.0',
+        '/usr/lib64/mutter-*',
+        '/usr/lib/mutter-*',
+        '/usr/lib/x86_64-linux-gnu/mutter-*',
+
+        './vala-girs/gir-1.0',
```
ewlsh, line 14: Was one of the benefits of this to get rid of `vala-girs`? Are there still GIRs we aren't hosting here?

examples/gjs/adw-1-hello/package.json
```diff
@@ -20,11 +20,11 @@
   "author": "Pascal Garber <pascal@artandcode.studio>",
   "license": "MIT",
   "devDependencies": {
-    "@babel/core": "^7.24.0",
+    "@babel/core": "^7.24.5",
     "@rollup/plugin-babel": "^6.0.4",
-    "rollup": "^4.13.0",
-    "typescript": "^5.4.2",
-    "vite": "^5.1.6"
+    "rollup": "^4.17.2",
```
ewlsh, line 25: In the future if we can separate version upgrades like this, it makes review faster

packages/cli/src/commands/copy.ts
```diff
@@ -0,0 +1,94 @@
+/**
```
ewlsh, line 1: Could we default to a list of directories based on XDG_DATA_DIRS and print what directories we are searching?

We could construct the shell and mutter directories based on the data directories too.

packages/lib/src/dependency-manager.ts
```diff
@@ -88,49 +125,241 @@ export class DependencyManager extends GirNSRegistry {
             : `import type ${namespace} from '${importPath}';`
     }
 
+    protected async parseGir(path: string) {
```
ewlsh, line 128: Caching the XML in my experience tends to lead to memory issues because you're caching massive strings for each library currently being processed. I don't think there is a need to cache the XML anymore, is there?

packages/lib/src/utils.ts
```diff
@@ -311,22 +312,76 @@ export const underscores = (str: string): string => {
...
+    return filesInfo
+}
+
+/**
+ * Find a file in a list of directories (sync)
+ * @param dirs The directories to search in
+ * @param filename The filename to search for
+ * @returns The file info
+ */
+export const findFilesInDirsSync = (dirs: string[], filename: string): FileInfo[] => {
```
ewlsh, line 354: It would be good to align on either using sync or async for file loading. We have several places with duplicate utilities for sync/async.

packages/lib/src/validators/interface.ts
```diff
@@ -10,7 +10,7 @@ export class InterfaceVisitor extends GirVisitor {
 
             if (!GObject) {
                 throw new Error(
-                    `GObject.Object could not be found while generating ${node.namespace.name}.${node.name}`
+                    `GObject.Object could not be found while generating ${node.namespace.namespace}.${node.name}`
```
ewlsh, line 13: I think the original was correct :thinking:

girs/Adw-1.gir
```diff
@@ -5151,6 +5151,9 @@ If the value is 0, the number of lines won't be limited.</doc>
         <type name="PreferencesRowClass" c:type="AdwPreferencesRowClass"/>
```
ewlsh, line 5151: GIRs as a submodule might make sense in the future. `types` we usually have to update in MRs, but `girs` does not have as tight a dependency.

ewlsh (changes requested):

.ts-for-gir.copy-all.rc.js
JumpLink: we should probably have all the girs ourselves now and could remove the submodule 👍

girs/Adw-1.gir
ewlsh: Alternatively I think you can flag files as "binary" to avoid them spamming MR review but I'm not sure that is ideal unless we have a CI check that validates the GIRs.

packages/lib/src/dependency-manager.ts
JumpLink: I have to parse the xml to extract the version number to compare the gir files for the copy command and so that I don't have to parse the xml again (before the real parsing starts) I cache it, but after that we could delete it again. Do you think that would make sense?

packages/lib/src/utils.ts
JumpLink: I actually wanted to stay with async but there was one place where that would have led to a lot of conversion work

girs/Adw-1.gir
JumpLink: yes, it makes sense to move them to a submodule, I can do that

packages/lib/src/validators/interface.ts
JumpLink: I'll have another look, I think I had renamed the proboerty

packages/lib/src/dependency-manager.ts
JumpLink: oh yes, I think it was in the ejs templates, but it may be that async now also works in ejs, I can try it out

packages/lib/src/utils.ts
JumpLink: oh yes, I think it was in the ejs templates, but it may be that async now also works in ejs, I can try it out

examples/gjs/adw-1-hello/package.json
JumpLink: okay

girs/Adw-1.gir
JumpLink: maybe the gir files are also interesting for other projects (like we used the ones from vala), so I think a submodule would be useful

packages/lib/src/validators/interface.ts
ewlsh: Sorry for the slow turn around, I find this renaming a bit confusing. `namespace.name` matches the pattern elsewhere.

packages/lib/src/utils.ts
ewlsh: Gotcha, might be worth testing async ejs in a future MR then.

packages/lib/src/dependency-manager.ts
ewlsh: I would cache the version number itself, caching the XML structure or parsed XML adds a lot of memory usage which will require increasing node's memory limits.

girs/Adw-1.gir
ewlsh: Would be good as a follow up

packages/lib/src/dependency-manager.ts
JumpLink: Okay, I will implement it like this

packages/lib/src/validators/interface.ts
ewlsh: I see this is aligning the previous properties in `GirModule` which calls the prop `namespace`, I still find it confusing but I don't think this is a blocking issue.

ewlsh (approved):
I've left comments, the only one I think is high priority is avoiding caching the XML. But that could be done in a follow up MR given the length of this one.

JumpLink:
@ewlsh That would be great, as I don't have much time at the moment, then this could already be merged and I will do that later

Issue #174, opened 2024-06-18
Create an import for the latest version in `*-ambient.d.ts`


vixalien:
For example Gtk generates a `gtk-4.0-ambient.d.ts`

```ts
declare module 'gi://Gtk?version=4.0' {
    const Gtk40: typeof import('./gtk-4.0.js').default;
    export default Gtk40;
}
```

Because this is the latest version, it would also be amazing to create the file with an additional import that doesn't include the `?version=4.0`

```ts
declare module 'gi://Gtk?version=4.0' {
    const Gtk40: typeof import('./gtk-4.0.js').default;
    export default Gtk40;
}

declare module 'gi://Gtk' {
    const Gtk40: typeof import('./gtk-4.0.js').default;
    export default Gtk40;
}
```

But on second thought, having this feature only for modules that don't have conflicting versions would be the better idea.

JumpLink:
@vixalien which version do you use? In `@girs/gtk-4.0 v4.14.3-4.0.0-beta.5` `gi://Gtk` is already exported without the version affix because it's the latest version. This is also the case for the other GObject libraries. Can you give it a try?

vixalien:
Oh nice. I just used the latest stable version with `npx ts-for-gir`.

I'll retry and update you!

JumpLink:
@vixalien With the latest version, all Ambient Modules are now generated without a version string. With the CLI argument `--onlyVersionPrefix` this can be deactivated, then only Ambient Modules with version strings are generated. Please give feedback if this works for you :)

Issue #173, opened 2024-06-18
Add `gi.d.ts` index file


vixalien:
Hi. Thanks for working on this!

I see there's now an option to generate an index `tsconfig.json`. Would it also be possible to generate a `gi.d.ts` file similar to https://gitlab.gnome.org/BrainBlasted/gi-typescript-definitions/-/blob/nightly/gi.d.ts?ref_type=heads where we can import just one file and have all the types available

JumpLink:
@vixalien Good idea. Would this also solve #168 for you? Maybe a part of it?

vixalien:
Most definitely. In fact the project almost meets the requirements.

I can theoretically do this manually by concatenating the `-ambient.d.ts` into a `gi.d.ts` file and then deleting them as they won't be needed 😅 but an option would be highly appreciated.

I would also delete the `-import.d.ts` files as they would be unused.

JumpLink:
@vixalien Is also now implemented with the latest version and is the default if `package.json` support is disabled. In addition to the `gi.d.ts` with all GIR types, an `index.d.ts` is now also generated. This means that in addition to all GIR types, all types including the GJS core types can be imported by a single file, now.

vixalien:
Wow, you're on a roll! I'll take a look ASAP

Issue #158, opened 2024-04-05
Finalise the ts-for-gir and gi.ts fusion


JumpLink:
Things we need to finish to get out of beta:

- [x] Replace the ts-for-gir parser with gi.ts/parser
- [x] CI tests should all be passed
- [X] Publish types on NPM
- [ ] Publish types on [JSR](https://jsr.io/)
- [X] Allow all versions to be imported without a version string (e.g. `gi://Gtk` instead of `gi://Gtk?version=4.0`), implemented in #176
- [X] Cleanup Code
- [x] Update documentation
- [x] #162
- [x] Refactor based on the review of  #162
- [x] Collect user feedback
- [x] Fix user problems

JumpLink:
@ewlsh I have made a refactor entry here so that it doesn't get lost

JumpLink:
I think most of the important work of this kind is now complete.

PR #176, opened 2024-07-15
Allow to toggle between package.json support + further cleanups from the gi.ts merge

### LINKED ISSUE #168: The `buildType` option doesn't work yet.
In the CLI, there's supposed to be an option called `buildType` to change the output format to just generate a .d.ts file, however that doesn't seem to be implemented yet.

Because of this, ts-for-gir can't really replace gi.ts right now

JumpLink:
* Renamed CLI argument `packageYarn` to `workspace` as other package managers such as [PNPM](https://pnpm.io/workspaces) also support this protocol
* Ambient module definitions are now always generated with and without versions prefix (previously this was only the case for the latest versions). I think this makes sense, as care should be taken not to use the same library in different versions anyway. If this is needed anyway, it can be prevented with the new generate option `onlyVersionPrefix`. This fixes `Allow all versions to be imported without a version string (e.g. gi://Gtk instead of gi://Gtk?version=4.0)` in #158 and possibly #174 
* A `index.d.ts` file are now created for each module with which everything for the module (ambient module, imports.gi and namespace) can be imported
* Synchronous methods removed that existed twice for the asynchronous variant, this was originally criticised in #162
* Support for type generation without `package.json` support to fix #168 and #173, for this there is a new cli agrument `package` to be able to toggle between `package.json` support. `package.json` support is disabled by default.


ewlsh:
Taking a look...

JumpLink:
@ewlsh I think for types without package.json support it is easier to generate all gir types in a single file, because in Ambient modules no relative imports are allowed and without package.json support only relative imports are possible. I will try that next

PR #177, opened 2024-07-17
gir: Promisify class members in typescript generator

ewlsh:
First stab at fixing promisifying...

`gi.ts` exposes a function called `promisifyFunctions` which handles creating promisified versions of functions based on their signature. Eventually this should instead use the async annotations GLib is working to support.

Needs testing to ensure output is correct


JumpLink:
@ewlsh looks like there are new conflicts, how are you handling conflicts so far?

JumpLink:
@ewlsh I have split `ModuleGenerator`, `NpmPackage` and `TypeDefinitionGenerator` into seperate files, this is the reason why you have a merge conflict. Do you want to fix it yourself or should I do it?

PR #184, opened 2024-07-22
 Preparations for GNOME Shell 46 and 47

JumpLink:
Includes a few small cleanups and preparations for GNOME Shell 46 and 47.


JumpLink:
I have reversed my fix attempt so that at least the rest can be merged

PR #201, opened 2024-09-05
Add GLib.Error.stack property

### LINKED ISSUE #196: GLib.Error.stack missing
GJS sets a `stack` property on all its errors, see https://gitlab.gnome.org/GNOME/gjs/-/blob/master/gi/gerror.cpp#L278

However, this property doesn't seem to exist on `GLib.Error` anymore, i.e.

```typescript
if (error instanceof GLib.Error) {
    console.info(error.stack);
}
```

no longer compiles.

It _did_ compile with types generated by the 3.x versions of ts-for-gir, so this looks like a regression.

swsnr:
Fixes #196

### DIFF
--- a/packages/lib/src/injections/glib.ts
+++ b/packages/lib/src/injections/glib.ts
@@ -19,6 +19,7 @@ import {
 } from "../gir.js";
 import { GirDirection } from "@gi.ts/parser";
 import { IntrospectedRecord } from "../gir/class.js";
+import { JSField } from "../gir/property.js";
 
 export default {
     namespace: "GLib",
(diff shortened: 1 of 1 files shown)


swsnr:
@JumpLink This is mostly an educated guess, from looking at other injection files.  Is this correct?  How do I test this locally?

swsnr:
Okay, I managed to run `yarn build:types:packages:gtk4` and now `GLib.Error` in `@types/glib-2.0` has a `stack` field:

GLib.Error

So presumably, this works?

swsnr:
@JumpLink Would you take a look?

JumpLink:
@swsnr Sorry for the late response, I'm still on parental leave at the moment. Looks good for me, thanks for the contribution!

PR #202, opened 2024-09-05
Improve System module types

### LINKED ISSUE #197: Improve types for `System` module
I was just verifying some types for `System` and both `addressOf` and `exit` seem like they have imprecise typings. I'm not sure if it is actually intended or not, so I'm opening this issue for discussion.

## Return type of `System.exit`

[`System.exit`](https://github.com/gjsify/ts-for-gir/blob/main/packages/generator-typescript/templates/gjs/system.d.ts#L148-L157) is marked with `void` instead of `never`. Consider using `never`, like `@types/node` uses for [`process.exit`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/process.d.ts#L884).

## Argument type of `System.addressOf`

This function is declared as `addressOf(o: any): string`, but `any` is a bit imprecise here (and almost anywhere). This function will always throw on non-object types, so the correct declaration should be `addressOf(o: object): string`. Keep in mind that TypeScript is aware of the difference between primitive types and their object versions (e.g. `string` vs `String`). You can test this in the REPL:

```js
gjs> const System = imports.system
gjs> System.addressOf({ prop: 'value' })
"0xa283ae3d240"
gjs> System.addressOf(12)
typein:14:8 Error: Error invoking addressOf, at argument 0 (object): Not an object
  @typein:14:8
  @<stdin>:1:42
gjs> System.addressOf('some text')
typein:16:8 Error: Error invoking addressOf, at argument 0 (object): Not an object
  @typein:16:8
  @<stdin>:1:42
gjs> System.addressOf(new String('some text'))
"0x300cf08144f0"
```

swsnr:
Closes #197

### DIFF
--- a/packages/generator-typescript/templates/gjs/system.d.ts
+++ b/packages/generator-typescript/templates/gjs/system.d.ts
@@ -48,7 +48,7 @@ export const programArgs: string[]
  * @param o Any Object
  * @returns A hexadecimal string (e.g. `0xb4f170f0`)
  */
-export function addressOf(o: any): string
+export function addressOf(o: object): string
 
 /**
  * Return the memory address of any GObject as a string.
@@ -141,7 +141,7 @@ export function dumpHeap(path?: string): void
  *   }
  * }
  * ```
- * 
+ *
  */
 export function dumpMemoryInfo(path?: string): void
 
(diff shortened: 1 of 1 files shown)


swsnr:
While I was at it, I figured implementing #197 was straight-forward to fix.

swsnr:
@JumpLink Would you take a look?

JumpLink:
@swsnr LGTM, thank you!

PR #225, opened 2024-11-13
Fix Generics for LayoutManager

### LINKED ISSUE #145: When constructing `St.Widget`, specifying `layout_manager` (as a subclass of `LayoutManager`) should be reflected in types
From https://github.com/gjsify/gnome-shell/issues/18

---

When one constructs a `St.Widget` like this:

```typescript
const myWidget = new St.Widget({
  layout_manager: new Clutter.GridLayout(),
})
```

The type of `myWidget.layout_manager` is still `LayoutManager`, which would make TypeScript errors when one tries to use methods from `GridLayout`.

## Suggestion 1: Multiple type parameters

```typescript
// St
class Widget<LayoutManagerT extends Clutter.LayoutManager | null = null, LabelActorT extends Clutter.LabelActor | null = null> extends Clutter.Actor<LayoutManagerT, LabelActorT> {
  constructor(config?: Widget.ConstructorProperties<LayoutManagerT, LabelActorT>)
}

// Clutter
interface Actor<LayoutManagerT extends LayoutManager | null, LabelActorT extends LabelActor | null> {
  layout_manager: LayoutManagerT
  layoutManager: LayoutManagerT
  label_actor: LabelActorT
  labelActor: LabelActorT
}
```

> [!WARNING]
> `ConstructorProperties` now shouldn't be a single interface, but a union of snake_case properties and camelCase properties. Keeping it as a single interface would have required user to specify both `layout_manager` and `layoutManager`, which would be absurd.

## Suggestion 2: Type dictionary as a single type parameter

```typescript
// St
class Widget<Config extends Widget.ConstructorProperties | null> extends ClutterActor<Config> {
  constructor(config: Config)
}

// Clutter
interface Actor<Config extends Widget.ConstructorProperties | null> {
  layout_manager: Config extends Widget.ConstructorProperties ? Config['layout_manager'] : null
  layoutManager: Config extends Widget.ConstructorProperties ? Config['layoutManager'] : null
  label_actor: Config extends Widget.ConstructorProperties ? Config['label_actor'] : null
  labelActor: Config extends Widget.ConstructorProperties ? Config['labelActor'] : null
}
```

> [!NOTE]
> I strongly recommend this solution. `ConstructorProperties` does not need to change at all.

---

The two suggestions above still have a flaw: It now requires TypeScript user to pass `null` explicitly, optional parameter is no longer allowed.

If you want to fix this flaw, then I suggest stop using the `class` keyword. Instead, you should utilize the TypeScript feature that allows merging variable and type, like so (assuming you choose "suggestion 2"):

```typescript
const Widget: unknown
  & (new <Config extends Widget.ConstructorProperties> (config: Config) => Widget<Config>)
  & (new (config?: null) => Widget<null>)
interface Widget<Config extends Widget.ConstructorProperties | null> extends Widget.Actor<Config> {}
```

As for me personally, whether I have to pass `null` explicitly makes little difference.

JumpLink:
## Description
Implements generic type support for St.Widget to provide proper type checking for layout managers. This change allows TypeScript to correctly infer and check layout manager types when using St.Widget with different layout managers.

## Related Issue
Fixes #145 

@KSXGitHub Can you try to test and review this changes?

## Type of Change
- [x] New feature (non-breaking change which adds functionality)
- [x] Type definition update

## Validation
The changes demonstrate proper generic type usage with St.Widget.

**Test Location:** `examples/st-15-layout-manager/main.ts`

**Test Case:**

```typescript
export class GridLayoutWidget extends St.Widget<Clutter.GridLayout> {
    static {
        GObject.registerClass({
            GTypeName: 'GridLayoutWidget',
        }, this);
    }
    constructor() {
        super({
            layout_manager: new Clutter.GridLayout()
        });
        // TypeScript now properly recognizes layout_manager methods
        this.layout_manager.attach(
            new St.Label({ text: 'Test' }),
            0, 0, 1, 1
        );
    }
}
```
**Expected Behavior:**
1. Type generator correctly adds generic support for St.Widget
2. TypeScript properly infers layout manager types
3. IDE provides correct type checking and autocompletion for layout manager methods

## Checklist
- [x] The implementation matches other generic type implementations (like Gtk)
- [x] Added tests to verify generic type generation


KSXGitHub:
What TypeScript definitions does this code generate?

JumpLink:
@KSXGitHub The key change in this PR adds generic type support for `St.Widget`, which generates TypeScript definitions that allow you to specify the type of layout manager when extending `St.Widget`. Here's what it means in practice:

```typescript
// Before the PR, St.Widget had no generic type parameter:
class Widget extends St.Widget {
    // layout_manager was typed as Clutter.LayoutManager
    layout_manager: Clutter.LayoutManager;
}

// After the PR, St.Widget accepts a generic type parameter:
class Widget<A extends Clutter.LayoutManager> extends St.Widget<A> {
    // layout_manager is now typed with the specific layout manager type
    layout_manager: A;
}
```

Looking at the example from the PR:

```typescript
export class GridLayoutWidget extends St.Widget<Clutter.GridLayout> {
    constructor() {
        super({
            layout_manager: new Clutter.GridLayout()
        });
        
        // TypeScript now knows that layout_manager is specifically Clutter.GridLayout
        // So these GridLayout-specific methods are properly typed:
        this.layout_manager.attach(
            new St.Label({ text: 'Test' }),
            0, 0, 1, 1  // column, row, column_span, row_span
        );
    }
}
```

The key benefits are:

1. TypeScript knows exactly which layout manager methods are available
2. You get proper type checking and autocompletion for the specific layout manager
3. TypeScript will catch errors if you try to use methods that don't exist on your chosen layout manager

For example, if you tried to use a `BoxLayout` method on a `GridLayout`, TypeScript would now catch that error at compile time.

KSXGitHub:
This solution is simpler than the ones I suggested, but this precisely is my concern:
1. This will only fix for `layout_manager`, does other properties (such as `label_actor`) have no subclasses?
2. Not defining a default type parameter would make this change a breaking change as the new definition would force the user to explicitly specify the `LayoutManager` subtypes, even when they only need `LayoutManager`.

JumpLink:
@KSXGitHub Sorry, I was too quick, I undid the changes again.

Maybe you would like to try and create a PR?

KSXGitHub:
I have no knowledge of the codebase though. Is the code generator capable of generating code as suggested by [my issue](https://github.com/gjsify/ts-for-gir/issues/145)?

PR #226, opened 2024-11-15
Define package properties as defined as soon as package.init was called

### LINKED ISSUE #220: `pkg.name` can be undefined
## Description

The current type of `pkg.name` can be undefined, but it's unclear how common that is. I have to use `pkg.name!` everywhere. This is also true for other types like `pkg.version`

## Current vs Expected Behavior

**Current:**
```typescript
    /** The base name of the entry point (eg. org.foo.Bar.App) */
    export const name: string | undefined;

```

**Expected:**
```typescript
    /** The base name of the entry point (eg. org.foo.Bar.App) */
    export const name: string;
```

## Example

```typescript
pkg.name
```

## References

- Documentation: 
- Source Code: 
- Related Issues: 
- GJS Version: 1.80.2

## Additional Context

- [ ] This affects other GJS types/classes
- [ ] This is a regression from a previous version
- [ ] This blocks other functionality

JumpLink:
## Description
Added improved TypeScript type definitions for package initialization to provide better type safety and developer experience. The changes include:

1. Added `UninitializedPackage` and `InitializedPackage` interfaces to represent package state
2. Updated type definitions to properly reflect the undefined/defined states before/after initialization
3. Enhanced `init()` function type to use type assertion for post-initialization state
4. Added example demonstrating TypeScript's type checking capabilities
5. Updated `pkg` global type to reflect initialization state

@vixalien Last problem with this solution is, that this works for `imports.package [props]` but not for `pgk[props]`, what do you think? Do you have any ideas how this could be done better? I could also create another import similar to `@girs/gjs/doc` but for `@girs/gjs/pkg-defined`.

## Related Issue
Fixes #220

## Type of Change
- [x] Type definition update
- [x] Documentation update

## Validation

**Test Location:** `examples/pkg-tsc/main.ts`

**Test Case:**

```ts
// Before initialization - TypeScript knows these are possibly undefined
console.log('name:', imports.package.name); // type: string | undefined

// Initialize the package
imports.package.init({
    name: 'org.example.App',
    version: '1.0.0',
    prefix: '/usr/local',
    libdir: '/usr/local/lib'
});

// After initialization - TypeScript knows these are definitely defined
console.log('name:', imports.package.name); // type: string

// TypeScript type check demonstration
function requiresInitializedPackage(pkgName: string) {
    return pkgName === imports.package.name; // No type error
}
```

**Expected Behavior:**
1. TypeScript should show proper type errors before initialization
2. After `init()` call, TypeScript should recognize all package properties as defined
3. Type checking should prevent accessing undefined properties
4. Runtime behavior remains unchanged

## Checklist
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have updated the documentation accordingly


JumpLink:
For the sake of simplicity, we will simply dispense with `undefined`, see #244

PR #241, opened 2025-02-26
Fix Object.get_property and Object.set_property

### LINKED ISSUE #190: Types for `GObject.Object.get_property` are wrong
In `ts-for-gir`, it's generated as 

```ts
get_property(property_name: string): any;
```

Instead of having an additional `GObject.Value` as defined here: https://gjs-docs.gnome.org/gobject20~2.0/gobject.object#method-get_property

JumpLink:
This fixes #190, does that look good for you @vixalien and @retrozinndev ?


retrozinndev:
Sorry for taking so long to answer. How can I build the types? I tried building it, but I keep getting workspace errors.

JumpLink:
@retrozinndev You need to clone this repo with the `--recurse-submodules` flag, or you can checkout the git submodule afterwards with `git submodule update --init` as described in [DEVELOPMENT.md](https://github.com/gjsify/ts-for-gir/blob/main/DEVELOPMENT.md). But if I don't know what kind of workspace error it is, I can only guess what it is.

PR #263, opened 2025-06-14
Experimental signal interfaces

JumpLink:
## Related Issue

Fixes #

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Type definition update
- [ ] Documentation update

## Validation

**Test Location:** `examples/[feature-name].ts`

**Test Case:**
```typescript
// Minimal test case demonstrating the changes
import Gio from 'gi://Gio';

// Example code
```

**Expected Behavior:**
1. Code should compile without type errors
2. Runtime behavior should match documentation/GJS implementation

## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] New and existing tests pass locally with my changes

### DIFF
--- a/examples/gtk-4-signal-interfaces/main.ts
+++ b/examples/gtk-4-signal-interfaces/main.ts
@@ -1,0 +1,183 @@
+/**
+ * GTK 4 Signal Interfaces Demo
+ * 
+ * This example demonstrates the new type-safe signal interfaces feature
+ * that enables TypeScript to infer signal names and callback signatures.
+ */
+
+import Gio from 'gi://Gio?version=2.0';
+import GLib from 'gi://GLib?version=2.0';
+import Gtk from 'gi://Gtk?version=4.0';
+
+// Initialize GTK
+Gtk.init()
+
+const loop = GLib.MainLoop.new(null, false)
+
+const app = new Gtk.Application({
+    application_id: 'com.github.gjsify.signal-interfaces-demo',
+    flags: Gio.ApplicationFlags.FLAGS_NONE,
+})
+
+app.connect('activate', () => {
+    // Create main window
+    const window = new Gtk.ApplicationWindow({ 
+        application: app,
+        title: 'Signal Interfaces Demo',
+        default_width: 600,
+        default_height: 400
+    })
+
+    // Create a header bar
+    const headerBar = new Gtk.HeaderBar()
+    window.set_titlebar(headerBar)
+
+    // Create main container
+    const box = new Gtk.Box({
+        orientation: Gtk.Orientation.VERTICAL,
+        spacing: 20,
+        margin_top: 20,
+        margin_bottom: 20,
+        margin_start: 20,
+        margin_end: 20,
+    })
+
+    // Title label
+    const titleLabel = new Gtk.Label({
+        label: '<span size="large" weight="bold">Type-Safe Signal Interfaces Demo</span>',
+        use_markup: true,
+        margin_bottom: 10,
+    })
+
+    // Description label
+    const descLabel = new Gtk.Label({
+        label: 'This demo shows how the new SignalSignatures feature enables\ntype-safe signal handling in TypeScript.',
+        justify: Gtk.Justification.CENTER,
+        margin_bottom: 20,
+    })
+
+    // Create buttons to demonstrate different signal types
+    const buttonBox = new Gtk.Box({
+        orientation: Gtk.Orientation.HORIZONTAL,
+        spacing: 10,
+        halign: Gtk.Align.CENTER,
+    })
+
+    // Button with simple signal (no parameters)
+    const simpleButton = new Gtk.Button({ label: 'Simple Signal' })
+    
+    // ✅ Type-safe: TypeScript knows this signal exists and takes no parameters
+    simpleButton.connect('clicked', () => {
+        console.log('Simple button clicked!')
+        statusLabel.set_text('Simple button clicked - no parameters needed')
+    })
+
+    // Button with complex signal handling
+    const complexButton = new Gtk.Button({ label: 'Complex Signal' })
+    
+    // ✅ Type-safe: TypeScript knows the exact signature
+    complexButton.connect('clicked', (_source: Gtk.Button) => {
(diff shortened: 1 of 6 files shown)


Aylur:
I could not test it, because `signalSignaturesSymbol` seem to be missing from the classes? so `GObject.SignalsOf` does not work.

1. Why did you decide to have the emitting object parameter fixed? It breaks subclasses

```ts
class MyButton extends Gtk.Button {}

const btn = new MyButton();
// src is still Gtk.Button but should be MyButton
btn.connect("activate", (src) => {});
```

It should be something like this imo

```ts

class Button {
  connect<K extends keyof Button.SignalSignatures>(
    signal: K,
    callback: (source: this, ...Parameters<Button.SignalSignatures[K]>): ReturnType<Button.SignalSignatures[K]>>
  ): number;

}

```

2. I appreciate the GObject type helpers, but imho they are a bit overkill. The `SignalSignatures`, `signalSignaturesSymbol` and `SignalsOf` should be sufficient

JumpLink:
> I could not test it, because `signalSignaturesSymbol` seem to be missing from the classes? so `GObject.SignalsOf` does not work.
> 
>     1. Why did you decide to have the emitting object parameter fixed? It breaks subclasses
> 
> 
> ```ts
> class MyButton extends Gtk.Button {}
> 
> const btn = new MyButton();
> // src is still Gtk.Button but should be MyButton
> btn.connect("activate", (src) => {});
> ```
> 
> It should be something like this imo
> 
> ```ts
> class Button {
>   connect<K extends keyof Button.SignalSignatures>(
>     signal: K,
>     callback: (source: this, ...Parameters<Button.SignalSignatures[K]>): ReturnType<Button.SignalSignatures[K]>>
>   ): number;
> 
> }
> ```
> 
>     2. I appreciate the GObject type helpers, but imho they are a bit overkill. The `SignalSignatures`, `signalSignaturesSymbol` and `SignalsOf` should be sufficient


That was not intended. The problem was that `this` caused it to refer to the signal interface and therefore no longer worked. I'll see if there's another way to solve this, but unfortunately I don't have much time to work on it.

JumpLink:
@Aylur Thanks for your honest review. I've now solved it with another helper, `OverrideFirstParameter`, to fix the problem with `this` . I don't know if there's a more elegant solution. The SignalsOf helper does not seem to be working properly yet either. If you can think of one, please let me know :)

Aylur:
mock of generated types

```ts
export namespace GObject {
  export type SignalCallback<Emitter, Fn> = Fn extends (...args: any[]) => any
    ? (source: Emitter, ...args: Parameters<Fn>) => ReturnType<Fn>
    : never;

  export namespace Object {
    // export interface Notify {
    //   (): void;
    // }

    export interface SignalSignatures {
      // honestly Object.Notify seems kind of useless since
      // its possible to index into SignalSignatures
      // so at this point why not just inline the declaration here?
      // notify: Notify;
      notify: (pspec: object) => void;
    }
  }

  export class Object {
    static $signals: Object.SignalSignatures;

    connect<K extends keyof Object.SignalSignatures>(
      signal: K,
      callback: SignalCallback<this, Object.SignalSignatures[K]>,
    ): number;

    connect(
      siganl: string,
      callback: (source: this, ...args: any[]) => any,
    ): number;
  }
}

export namespace Gtk {
  export namespace Widget {
    export interface SignalSignatures extends GObject.Object.SignalSignatures {
      "state-flags-changed": (flags: number) => void;
    }
  }

  export class Widget extends Object {
    static $signals: Widget.SignalSignatures;

    connect<K extends keyof Widget.SignalSignatures>(
      signal: K,
      callback: GObject.SignalCallback<this, Widget.SignalSignatures[K]>,
    ): number;

    connect(
      siganl: string,
      callback: (source: this, ...args: any[]) => any,
    ): number;
  }

  export namespace Button {
    export interface SignalSignatures extends Widget.SignalSignatures {
      clicked: () => void;
    }
  }

  export class Button extends Widget {
    static $signals: Button.SignalSignatures;

    connect<K extends keyof Button.SignalSignatures>(
      signal: K,
      callback: GObject.SignalCallback<this, Button.SignalSignatures[K]>,
    ): number;

    connect(
      siganl: string,
      callback: (source: this, ...args: any[]) => any,
    ): number;
  }
}
```

user code:

```ts
import { GObject, Gtk } from "./gjsify-mock.ts";

interface MyButtonSignals extends Gtk.Button.SignalSignatures {
  "my-signal": (arg: number) => number;
}

class MyButton extends Gtk.Button {
  declare static $signals: MyButtonSignals;

  override connect<K extends keyof MyButtonSignals>(
    signal: K,
    callback: GObject.SignalCallback<this, MyButtonSignals[K]>,
  ): number {
    return super.connect(signal, callback);
  }
}

const a = new MyButton();
// works
a.connect("state-flags-changed", (source, flags) => {});

declare function jsx<
  T extends typeof GObject.Object,
  Sig extends keyof T["$signals"],
  Callback extends T["$signals"][Sig],
>(
  obj: T,
  signal: Sig,
  callback: GObject.SignalCallback<InstanceType<T>, Callback>,
): number;

// works and this is exactly my use case
jsx(MyButton, "state-flags-changed", (source, pspec) => {});

// there is no way to make it work afaik
// since we can't reference the constructor type of an instance type
// as a workaround we could define both static and instance $signals annotations
jsx(a, "notify", (source, pspec) => {});
```

I think the the signal annotations (`$signals`) should be public on classes so that subclasses can define their signal types, although in user code it might be a lot of boiler plate as for signals to work at runtime and be annotated a `GObject.registerClass`, `connect` override and `$signals` annotation is needed but I can't come up with a better way to make this work.

Also, the `GObject.SignalCallback` is just for this example to be less lines, I don't actually think that it should be exported and used in user code. In user code `Parameters` and `ReturnType` should be preferred or users can define this helper themselves.

JumpLink:
@Aylur It should now be as you suggested. Can you check that again?

PR #287, opened 2025-08-24
revert variant parsing changes

### LINKED ISSUE #286: [Regression] Variant parser fails on nested dictionaries
## Description
With the original parser code from @ewlsh this used to be parsed correctly:

```ts
// { [key: string]: { [key: string]: unknown } }
type NestedDict = $ParseRecursiveVariant<"a{sa{sv}}">
```

With the changes introduced in #279 it now fails with an error.

Another change I failed to notice from the PR is that `v` type (aka `GLib.Variant<any>`) unpacks to `any` where in the original code it used to unpack to `unknown` which I feel like was more appropriate.

Aylur:
Simply copied the code from before #279, from the previous template https://github.com/gjsify/ts-for-gir/blob/e1bfc49d233b4f13d4a3b342d5fc636532c4f352/packages/lib/src/generators/dts/glib.ts

I'm not entirely sure why you introduced those changes in #279, and the actual code is way beyond me, so I'm not even going to try fixing it, reverting works.

closes  #286

### DIFF
--- a/packages/templates/templates/glib-2.0.d.ts
+++ b/packages/templates/templates/glib-2.0.d.ts
@@ -13,448 +13,327 @@
 // Utility types for variant parsing
 type VariantTypeError<T extends string> = { error: true } & T;
 
-// === Core parsing utilities ===
-
-/**
- * Maps basic GVariant type characters to TypeScript types
- */
-type BasicTypeMap<T extends string> =
-    T extends 'b' ? boolean :
-    T extends 's' | 'o' | 'g' ? string :
-    T extends 'n' | 'q' | 't' | 'd' | 'u' | 'i' | 'x' | 'y' ? number :
-    T extends 'h' | '?' ? unknown :
-    T extends 'v' ? Variant :
-    never;
-
 /**
  * Creates index type for dictionaries based on key type
  */
-type CreateIndexType<Key extends string, Value extends any> =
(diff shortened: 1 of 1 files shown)


Aylur:
Most failing tests are now either because the test case is wrong or because unpacking now a variant results in `unknown` where an `any` was expected


For example even the first one is wrong
```ts
type IntTuple = Infer<"(ii)">; // Should be [Variant<any>, Variant<any>]
```
It should be `Variant<"i">` and not `Variant<any>`, this one only tests for `Variant.*` so it will pass, but almost every test "description" is wrong though, but they are passing because they only test for `Variant.*`.

The ones that fail are the ones where `any` is expected, but the parser returns `{ [key: string]: unknown }` which is actually correct.

Aylur:
@JumpLink  I'll happily update the tests, but before I do: Is there a specific reason to use a language server like this and using regexes to test types? wouldn't it be more appropriate and easier to write/maintain regular typescript files and using an `Equal<T, Expected>` utility and `tsc` or maybe something like the `ts-expect` package? It seems a bit unnecessary to test variants at runtime. If you do want to also test runtime, how about using `tsc` + [jasmine-gjs](<https://gitlab.gnome.org/GNOME/gjs/-/blob/master/doc/Testing.md?ref_type=heads>)?

PR #318, opened 2026-03-05
Add JSR registry support and normalize package naming

JumpLink:
This PR adds [JSR (JavaScript Registry)](https://jsr.io) support to ts-for-gir and normalizes package naming across both NPM and JSR.

**Breaking Change:** All `@girs/*` NPM package names now use hyphens instead of dots (e.g., `@girs/gtk-4-0` instead of `@girs/gtk-4.0`). This affects all imports and dependencies.

### Changes

- **Normalized package names**: Dots replaced with hyphens in all package names. JSR doesn't allow dots in package names, and rather than maintaining two naming schemes, we unified both NPM and JSR under the same convention.
- **JSR support**: Each generated package now includes a `jsr.json` and a `mod.ts` entrypoint for publishing to JSR.
- **`mod.ts` entrypoint**: Uses `globalThis.imports.gi` instead of the `gi://` protocol, since JSR only allows `jsr:`, `npm:`, `node:`, and `data:` import specifiers.
- **Renamed template files**: e.g., `gobject-2.0.d.ts` → `gobject-2-0.d.ts` to match the new naming convention.
- **Updated all examples**: Imports, `package.json` dependencies, and `tsconfig.json` paths adjusted to use new package names.

### What's not included yet

- The `@girs/gjs` package needs special handling for JSR due to `declare global` blocks and will reference the NPM package as a dependency instead.

### Test plan

- [ ] Verify `npx jsr publish` works for generated packages
- [ ] Verify examples still build with the renamed packages
- [ ] Verify generated type definitions are complete (e.g., `registerClass` in GObject)


JumpLink:
This is being closed because JSR is primarily intended for executable TypeScript, not for pure type declarations (`.d.ts`) packages.

PR #357, opened 2026-04-11
Revert "Mark enum parameters as nullable, fixes #207"

### LINKED ISSUE #207: enum parameters should accept null.
Say, for example `Soup.Server.listen_local`

the `options` parameter is an enum, but it should accept `null` as GJS treats that as a `0`.

ptomato:
## Description

This reverts commit a8490ef57e7ed5d9f3be999b6c42b9c27d1ace5e. See https://github.com/gjsify/ts-for-gir/issues/207

This is technically correct, but not desirable in type definitions. null is allowed for enum params, but a lot of the GJS marshalling stuff dates from a time when TS didn't exist and JS was a lot more lax about argument coercion.

For example: Math.PI.toFixed('3') works, but isn't correct according to TS type definitions.

(Not sure if you want to keep the `OptionsBase`/`OptionsGeneration` change and just remove the enum params visitor; that would be an alternative to this PR)

## Related Issue

- See #207
- Reverts #213

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)

## Validation

Existing tests pass, regenerating `types-dev` only removes `| null` from enum params

**Expected Behavior:**
1. Code should compile without type errors
2. Runtime behavior should match documentation/GJS implementation

## Checklist


JumpLink:
Thank you @ptomato 👍

PR #380, opened 2026-05-06
Make raw pointer (`void*`) handling more like GJS

ptomato:
## Description

GJS does not handle raw pointers (i.e., `void*` in C).

If a function takes a nullable raw pointer, null is the only accepted value you can pass in JS. If a function takes a non-nullable raw pointer, it's impossible to call it — it will always throw no matter what value you pass.

These behaviours can be represented in TypeScript with the types `null` and `never`, respectively.

Return values are slightly different. Functions that return raw pointers will always return null in JS, even if the C function didn't return null and even if not nullable.

I thought about a few alternatives, as well. One alternative could be to make the TS return type of a function `never` if it takes but does not return a non-nullable raw pointer, to signify that it can never return. Another approach might be to delete such functions from the type declaration altogether.

## Related Issue

N/A

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Type definition update
- [ ] Documentation update

## Validation

**Test Location:** `examples/glib-2-types/main.ts`

## Checklist

- [x] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/examples/glib-2-types/main.ts
+++ b/examples/glib-2-types/main.ts
@@ -313,6 +313,31 @@ function testNullableParamsRequired(): void {
 	}
 }
 
+function testRawPointers(): void {
+	// Non-nullable raw pointers cannot be passed to functions, either according
+	// to TypeScript or at runtime
+	let failed = false;
+	try {
+		// @ts-expect-error
+		GLib.str_hash("foobar");
+	} catch (_) {
+		failed = true;
+	}
+	if (!failed) throw new Error("Passing value to non-nullable raw pointer failed");
+
+	// Nullable raw pointers are always marshalled as C null, even if you pass
+	// something else
+	// @ts-expect-error
+	const nonnullHash = GLib.direct_hash("foobar");
+	const nullHash = GLib.direct_hash(null);
+	if (nonnullHash !== nullHash) throw new Error("Passing value to nullable raw pointer failed");
+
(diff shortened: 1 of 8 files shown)


JumpLink:
@ptomato That looks great, thank you very much!

My LLM has pointed out the following lines to me which may allow `RawPointerType/AnyType` through:
```
  - packages/lib/src/gir/property.ts:213 — new NullableType(property.type)                                                                                                                                                                                                                 
  - packages/lib/src/gir/introspected-classes.ts:80 — new NullableType(prop.type)
  - packages/lib/src/injections/gobject.ts:614 — new NullableType(param.type)                                                                                                                                                                                                              
  - packages/lib/src/injections/shell.ts:27 — wrappt OrType/TypeIdentifier    
``` 

Could you check that?

ptomato:
Thanks. I took a look at each of those sites - looks like the first two were added last week, so I didn't catch them when I did a search for NullableType almost a month ago :sweat_smile: 
The 3rd site seems to only pass `GObject.Closure` to NullableType and the 4th only passes `St.SnippetHook | Cogl.SnippetHook`, so they can't be `any` or `never`.

The PR is updated now.

JumpLink:
@ptomato Thank you very much :)

PR #391, opened 2026-05-10
fix: Output formatting improvements

### LINKED ISSUE #379: Prettier output was better for union types
## Description

Dropping the prettier output in #376 led to some weird-looking union types. Reading the rationale there, I guess the idea is that the code generator should produce good-looking output without prettier, so would you be interested in some patches to make the output more like the previous prettier-processed output?

## Current Behavior

Union types are output with parentheses, and we also get some particularly bad-looking ones such as:

```typescript
function prefix_error_literal(err: ((Error | null) | null), prefix: string): (Error | null) | null;
```

We also get some inconsistent comma/semicolon use and double/single quotes:
```typescript
constructor(options: { message: string, code: number });
static "2BIG": number;
static ["new"](): AsyncQueue;
```

## Expected Behavior

```typescript
function prefix_error_literal(err: Error | null, prefix: string): Error | null;
constructor(options: { message: string; code: number });
static '2BIG': number;
static ['new'](): AsyncQueue;
```

## Steps to Reproduce (for bugs)

N/A

## Environment

N/A

## Additional Context

- [ ] This is blocking my work
- [x] I would like to help implement this feature
- [ ] I have a workaround (please describe below)

ptomato:
## Description

Three commits to make the output formatting of the .d.ts files nicer.

- Format record passed to GError constructor with `;`
- Avoid empty bodies on separate lines (output `{}` instead of `{\n\n}`
- Avoid extra parentheses in union types

I didn't actually fix the single vs double quotes that I mentioned in the issue, because I realized they are consistent after all; just consistently the opposite of what prettier rendered. So I left that as-is.

## Related Issue

Fixes #379

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)

## Validation

Validated by inspection of GLib-2.0.d.ts, GObject-2.0.d.ts, Gio-2.0.d.ts and running the test suite.

## Checklist

### DIFF
--- a/packages/generator-typescript/src/generators/signal-generator.ts
+++ b/packages/generator-typescript/src/generators/signal-generator.ts
@@ -84,14 +84,14 @@ export class SignalGenerator {
 
 		parentSignatures.push(...interfaceSignatures);
 
-		let signatureHead: string;
+		let signatureDecl: string;
 		if (parentSignatures.length > 0) {
-			signatureHead = `${indent}interface SignalSignatures extends ${parentSignatures.join(", ")} {`;
+			signatureDecl = `interface SignalSignatures extends ${parentSignatures.join(", ")}`;
 		} else {
 			const isGObjectObject = girClass.name === "Object" && girClass.namespace.namespace === "GObject";
 
 			if (isGObjectObject) {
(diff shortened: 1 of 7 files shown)


JumpLink:
@ptomato thank you!

PR #435, opened 2026-08-04
fix(cli): propagate the dev wrapper's exit code

JumpLink:
`bin/ts-for-gir-dev` ended on a bare `spawn(...)` — no `exit` handler, no `error` handler, 43 lines total. The wrapper returned immediately, so **the process exited 0 no matter what the CLI did.**

Measured before the fix, driving the real wrapper against a stub child:

| child | before | after |
|---|---|---|
| exits 0 | 0 | 0 |
| exits 1 | **0** | 1 |
| exits 7 | **0** | 7 |
| SIGTERM | **0** | 143 |

And it was not only subcommand failures. In a tree with no `node_modules` the CLI crashed in `node:internal/modules/package_json_reader` before executing a line — a hard startup failure, reported as success.

## Why it matters

`build:types` is

```
ts-for-gir-dev generate --configName=… --girDirectories=./girs … && gjsify install
```

so the step that generates all **~700 `@girs/*` packages could not fail in CI.** A generator crash, an unresolvable dependency, a thrown parse error — every one landed as a green step, leaving a human reading the log as the only signal.

## The fix

The child's outcome becomes the wrapper's:

- the **exact** code, not merely non-zero — collapsing everything to 1 would lose a CLI that distinguishes its codes, and a caller keying on one would silently take the wrong branch;
- a signal **re-raised on ourselves**, so a killed child surfaces as the conventional 128+n rather than 0;
- `error` handled too, because a failure to spawn at all was the same silence;
- `process.execPath` instead of resolving `node` from PATH again — this wrapper already runs under a Node that accepts the experimental flags it passes, and a second PATH lookup can find a different one.

## The test

`tests/e2e/dev-wrapper-exit/` copies the **real** wrapper into a temp tree beside a stub `src/start.ts` whose exit code comes from argv. The wrapper resolves its target as `../src/start.ts` relative to its own location, so the copy drives the real file with a controllable child. That makes the exit code an *input* rather than something to provoke, and the suite needs no dependency tree, no GIR files and no generation run — provoking a real failure would test the CLI, this tests the wrapper.

It runs **first** in the e2e chain: about a second, no build required, and a broken wrapper invalidates every later suite that shells through it.

```
✔ passes a successful run through as 0
✔ fails when the child fails
✔ preserves the exact code, not just non-zero
✔ reports a signal-killed child as 128+n rather than success
✔ fails loudly when the interpreter cannot be spawned at all
  5 pass, 0 fail
```

### DIFF
--- a/packages/cli/bin/ts-for-gir-dev
+++ b/packages/cli/bin/ts-for-gir-dev
@@ -39,5 +39,36 @@ const nodeArgs = [
 	...process.argv.slice(2), // Forward all CLI arguments to the TypeScript file
 ];
 
-// Spawn the Node.js process with TypeScript support and inherit stdio
-spawn("node", nodeArgs, { stdio: "inherit" });
+// Spawn the Node.js process with TypeScript support and inherit stdio.
+//
+// `process.execPath`, not the string "node": this wrapper already runs under a Node
+// that accepts the experimental flags above, and resolving `node` from PATH again can
+// pick a different one.
+const child = spawn(process.execPath, nodeArgs, { stdio: "inherit" });
+
+// PROPAGATE THE CHILD'S OUTCOME. Without this the wrapper returned immediately and the
+// process exited 0 no matter what the CLI did — so `ts-for-gir-dev generate …` reported
+// SUCCESS on a failed generation. Not cosmetic: the repo's own `build:types` is
+//
+//     ts-for-gir-dev generate --configName=… && gjsify install
+//
+// so the step that generates all ~700 `@girs/*` packages could not fail in CI. A
+// generator crash, an unresolvable dependency, a thrown parse error — every one of them
+// landed as a green step, leaving a human reading the log as the only signal.
+child.on("error", (error) => {
+	console.error(`ts-for-gir-dev: failed to spawn ${process.execPath}: ${error.message}`);
+	process.exit(1);
+});
+
+child.on("exit", (code, signal) => {
+	// A signal-killed child must not report success either. Re-raise the same signal on
+	// ourselves so the caller sees the real cause (128+n) instead of a synthetic code —
+	// `default` is restored first, otherwise the re-raise is swallowed by our own handler
+	// for signals Node installs one for (SIGINT/SIGTERM).
+	if (signal) {
+		process.removeAllListeners(signal);
+		process.kill(process.pid, signal);
+		return;
+	}
+	process.exit(code ?? 1);
+});
--- a/tests/e2e/dev-wrapper-exit/run.mjs
+++ b/tests/e2e/dev-wrapper-exit/run.mjs
@@ -1,0 +1,119 @@
+// E2E test for `packages/cli/bin/ts-for-gir-dev` — the dev entry's exit-code
+// propagation.
+//
+// THE INCIDENT
+//
+// The wrapper spawns `src/start.ts` under Node's type-stripping flags and used to end
+// on a bare `spawn(...)`: no `exit` handler, no `error` handler, 43 lines total. So the
+// wrapper returned immediately and the process exited **0 no matter what the CLI did**.
+//
+// Measured before the fix: a bogus subcommand exited 0, and so did a run where the CLI
+// could not even load its own dependencies (`node:internal/modules/package_json_reader`
+// crash in a tree with no `node_modules`) — a hard startup failure reported as success.
+//
+// It mattered because of one script. The repo's `build:types` is
+//
+//     ts-for-gir-dev generate --configName=… --girDirectories=./girs … && gjsify install
+//
+// so the step that generates all ~700 `@girs/*` packages COULD NOT FAIL in CI. A
+// generator crash, an unresolvable dependency, a thrown parse error: every one of them
+// landed as a green step, and the only remaining signal was a human reading the log.
+//
+// WHY THE TEST DRIVES A STUB
+//
+// It copies the REAL wrapper into a temp tree beside a stub `src/start.ts` whose exit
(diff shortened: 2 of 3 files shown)


JumpLink:
**The two red checks are not this PR — they are gjsify 0.28.0, and this run is the first one to hit it.**

`ci.yml:68` bootstraps `https://github.com/gjsify/gjsify/releases/latest/download/install.mjs`, so every run installs whatever gjsify's *latest* release is at that moment:

| | |
|---|---|
| gjsify **v0.27.1** published | 2026-08-04 **06:28:50Z** |
| ts-for-gir main last **green** | 2026-08-04 **08:32:59Z** ← already on 0.27.1 |
| gjsify **v0.28.0** published | 2026-08-04 **17:03:45Z** |
| this PR's run | 2026-08-04 **19:49Z** |

The only variable between the green run and this one is 0.27.1 → 0.28.0. And neither failing command can be reached by this diff, which touches `bin/ts-for-gir-dev`, a `tests/e2e` script entry, and one new test directory.

**`check` — `gjsify lint` re-executes the gjsify bundle instead of oxlint:**

```
Running on GJS 1.86.0 (SpiderMonkey)
script "check" exited with code 1
```

That is the documented `process.execPath`-under-GJS trap: under the committed GJS bundle `process.execPath` *is* the CLI, so spawning it re-runs gjsify with oxlint's arguments and yargs prints top-level help. gjsify's AGENTS.md describes this as fixed (`nodeBinary()` instead of `process.execPath`) and guarded by `tests/e2e/lint-gjs-launcher` — so on a host with no `node`, that guard is not covering the path this CI takes.

**`test-e2e` — the bundler engine is not loadable on the same host:**

```
gjsify build: no usable bundler engine under GJS — `@gjsify/rolldown-native` is not
loadable (its JS facade lib/ is not built, or its prebuild typelib is missing from
GI_TYPELIB_PATH)
```

A global GJS install is supposed to lay the engines down itself (`installGjsEnginePackages()`), since there is no npm fallback there.

Both are the **Node-free GJS consumer path**, which is exactly what this CI exists to exercise — so it is doing its job as a canary. Filed on the gjsify side; not fixing it here.

Separately worth noting because it is independent of any version bug: pulling `releases/latest` means **a CI result here depends on the time of day, not only on the commit.** The declared pin is `@gjsify/cli ^0.14.0`, fourteen minors away from what the bootstrap actually installs — the two disagree and the bootstrap wins.

JumpLink:
**Correction to my comment above: the "host with no `node`" attribution is wrong.** I measured it afterwards and it does not hold.

On a genuinely node-less host, gjsify 0.28.0's `gjsify lint` does exactly the right thing — reproduced locally by running the 0.28.0 GJS bundle under `gjs` with a full userland minus `node`/`npm`/`npx`:

```
[gjsify oxc] cannot run oxlint: no `node` binary on PATH.
  oxlint IS installed — its launcher was found. What is missing is the
  interpreter that runs it: oxlint's launcher is a Node ESM script.
```

Clean, honest, exit 1. So the no-node path is **not** the defect, and the `tests/e2e/lint-gjs-launcher` guard is not the gap I claimed it was.

**What still holds, and is all I can support with evidence:**

1. **The bisect boundary**, from release timestamps against this repo's own run history: 0.27.1 (06:28:50Z) is good — main was green at 08:32:59Z already on it; 0.28.0 (17:03:45Z) is bad — this run at 19:49Z is the first after it.
2. **The symptom.** `Unknown arguments: config, …/node_modules/oxlint/bin/oxlint, .` is gjsify's yargs receiving **oxlint's** argv — the launcher path as a positional, `--config` and `.` as flags. That is the CLI having spawned *itself* instead of a Node interpreter, which is the `process.execPath`-under-GJS shape.
3. **This PR's diff cannot reach either failing command** (`gjsify run check`, `gjsify run build:app`); it touches `bin/ts-for-gir-dev`, one `tests/e2e` script entry, and a new test directory.

What I do **not** yet know is which condition on this runner selects the self-spawn, since a node-less host does not and a node-present host does not either (both measured locally). So the trigger is something else about this environment — most likely how the globally installed CLI resolves the interpreter there.

Leaving this PR red rather than merging on a green-looking rebase: the failures are real, they are just not this change's.

JumpLink:
## The red `check`/`test-e2e` here is #437's, not this PR's

Diagnosed rather than assumed. Both failing runs predate #437 (this one: see the run dates), and `check` fails with the verbatim signature #437 documents:

```
```

That is the `@gjsify/cli@0.14.0` self-re-execution bug. `ci.yml` puts `node_modules/.bin` **first** on PATH, and `main`'s `gjsify-lock.json` still pins `@gjsify/cli: 0.14.0` — so the bootstrapped CLI loses the PATH lookup to a 16-versions-stale one. 0.14.0 predates gjsify `78ae67790` (spawn oxlint via `process.execPath`), so under GJS the CLI re-executed **itself** with oxlint's argv. `test-e2e` fails for the sibling reason: no bundler engine (gjsify#1005).

**Nothing to fix on this branch.** #437 is the fix, and it is now retargeted to `@gjsify/cli` **0.30.0** — 0.29.0 could not have worked either, because `test:cli-tarball-shape` shells out to `gjsify pack`, which never exits under GJS on a package declaring `prepack` (gjsify#1010, measured at 5m30s against two independent 5-minute ceilings). gjsify#1011 fixed it and landed *between* the v0.29.0 and v0.30.0 tags. Verified locally on #437's branch: that suite now passes 8/8 in **6.4 s**.

So the order in gjsify ADR 0019 § Implementation — **#437 → #435 → #436** — is not a preference, it is what these logs say. This PR wants a rebase onto `main` once #437 lands, which re-runs CI against the 0.30.0 pin.

PR #436, opened 2026-08-05
fix(lib): only claim a version the library states

JumpLink:
`libraryVersion` is published on every `@girs/*` package and named in gjsify's ADR 0008 — and for a long time nothing read it. Once `@gjsify/cli`'s `system-check` began comparing it against the *installed* library, the field turned out to mean two different things with no way to tell them apart.

## Measured across 32 published `@girs/*` packages

| | |
|---|---|
| real upstream release | **12** (gtk-4.0 → 4.23.0, adw-1 → 1.10.0, glib/gio/gobject → 2.88.0, …) |
| the **namespace** version padded to three components | **17** (gdk-4.0 → `4.0.0`, cairo-1.0 → `1.0.0`, gsk-4.0 → `4.0.0`, …) |
| absent | 3 |

GDK ships *inside* GTK and states no version of its own, so `@girs/gdk-4.0` reads `4.0.0` — a string shaped exactly like a release that corresponds to none. Compared against a host GTK 4.22.4 that reports an 18-minor skew which does not exist, so the consumer had to carry a hand-maintained list of which packages to believe. **A value that must be second-guessed per package is worse than no value.**

## What changes

`LibraryVersion` records whether the library **itself** stated the version, via `MAJOR_VERSION`/`MINOR_VERSION`/`MICRO_VERSION` (or the `VERSION_*` spelling), and the package template omits the field when it did not. Absence is a fact a consumer can act on.

`PACKAGE_DESC` stops making the claim too: every `@girs/gdk-4.0` published so far said *"generated from library version 4.0.0"*, which was not true of the GIR it was generated from.

## An operator-precedence bug, found while reading it

```js
if (name === "MAJOR_VERSION" || (name === "VERSION_MAJOR" && value)) {
    this.major = Number(value) || undefined;
```

`&& value` binds to the **second** spelling only. A valueless `MAJOR_VERSION` constant therefore took the first branch and assigned `Number(undefined) || undefined`, **erasing** the component the namespace fallback had just supplied — turning `4.0` into `0.0.0` while looking like a version bump. Hoisted so the guard covers both spellings; pinned by a test.

Left alone deliberately: `Number("0") || undefined` coerces a legitimate zero component to `undefined`. `toString()`'s `|| "0"` masks it, and changing it would move `compare()`, which picks the newest GIR among duplicate candidates.

## Tests, and where they live

`tests/library-version/` — 10 cases: declared, `VERSION_*` spelling, namespace-only, bare namespace, the valueless-constant regression, partial declaration, ordering, and the three `PACKAGE_DESC` shapes. All verified locally against the real source.

They are **not** beside the source, because that is where this repo's runner looks: CI runs `test:tests` = `gjsify foreach test --include '@ts-for-gir-test/*'`, so a vitest suite under `packages/**` never executes. `packages/language-server/src/__tests__` is in exactly that position today — worth a separate look.

## Note on CI

A generation regression from this change would currently be **invisible** in CI: `build:types` runs through `bin/ts-for-gir-dev`, which discards its child's exit code (#435). That PR is the prerequisite for trusting a green run here.

### DIFF
--- a/packages/lib/src/constants.ts
+++ b/packages/lib/src/constants.ts
@@ -38,7 +38,11 @@ export const APP_SOURCE = "https://github.com/gjsify/ts-for-gir";
 export const APP_VERSION = getPackageVersion();
 
 export const PACKAGE_DESC = (packageName: string, libraryVersion?: LibraryVersion) => {
-	if (libraryVersion) {
+	// "generated from library version X" is a CLAIM, so it is only made when the library
+	// stated X itself. For a namespace-derived version it was false in every published
+	// description: `@girs/gdk-4.0` read "generated from library version 4.0.0" while the
+	// GIR it was generated from came with GTK 4.2x. See LibraryVersion.declaredByLibrary.
+	if (libraryVersion?.declaredByLibrary) {
 		return `GJS TypeScript type definitions for ${packageName}, generated from library version ${libraryVersion.toString()}`;
 	}
 	return `GJS TypeScript type definitions for ${packageName}`;
--- a/packages/lib/src/library-version.ts
+++ b/packages/lib/src/library-version.ts
@@ -5,6 +5,28 @@ export class LibraryVersion {
 	minor: number | undefined;
 	patch: number | undefined;
 
+	/**
+	 * Whether the library ITSELF stated this version, through `MAJOR_VERSION` /
+	 * `MINOR_VERSION` / `MICRO_VERSION` (or the `VERSION_*` spelling) constants in its
+	 * GIR — as opposed to it being derived from the NAMESPACE version in the filename.
+	 *
+	 * The distinction is the entire point, because the two render identically and mean
+	 * completely different things. `Gtk-4.0.gir` declares 4.23.0. `Gdk-4.0.gir` declares
+	 * nothing — GDK ships INSIDE GTK and has no version of its own — so the namespace
+	 * fallback below renders `4.0.0`, a string shaped exactly like a release that
+	 * corresponds to no release. Measured across 32 published `@girs/*` packages: 12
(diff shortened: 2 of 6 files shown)


JumpLink:
## The red `check`/`test-e2e` here is #437's, not this PR's

Diagnosed rather than assumed. Both failing runs predate #437 (this one: see the run dates), and `check` fails with the verbatim signature #437 documents:

```
```

That is the `@gjsify/cli@0.14.0` self-re-execution bug. `ci.yml` puts `node_modules/.bin` **first** on PATH, and `main`'s `gjsify-lock.json` still pins `@gjsify/cli: 0.14.0` — so the bootstrapped CLI loses the PATH lookup to a 16-versions-stale one. 0.14.0 predates gjsify `78ae67790` (spawn oxlint via `process.execPath`), so under GJS the CLI re-executed **itself** with oxlint's argv. `test-e2e` fails for the sibling reason: no bundler engine (gjsify#1005).

**Nothing to fix on this branch.** #437 is the fix, and it is now retargeted to `@gjsify/cli` **0.30.0** — 0.29.0 could not have worked either, because `test:cli-tarball-shape` shells out to `gjsify pack`, which never exits under GJS on a package declaring `prepack` (gjsify#1010, measured at 5m30s against two independent 5-minute ceilings). gjsify#1011 fixed it and landed *between* the v0.29.0 and v0.30.0 tags. Verified locally on #437's branch: that suite now passes 8/8 in **6.4 s**.

So the order in gjsify ADR 0019 § Implementation — **#437 → #435 → #436** — is not a preference, it is what these logs say. This PR wants a rebase onto `main` once #437 lands, which re-runs CI against the 0.30.0 pin.

PR #437, opened 2026-08-05
fix(ci): bump @gjsify/cli to 0.30.0 and install json-glib

JumpLink:
`check` and `test-e2e` have been red. There are **two independent causes**, both fixed here, both measured.

## 1. The CLI that ran was 0.14.0, not the one we bootstrap

`ci.yml` bootstraps `@gjsify/cli` from `releases/latest`, then runs:

```yaml
- run: PATH="$PWD/node_modules/.bin:$PATH" gjsify run check
```

`node_modules/.bin` comes **first**, and `gjsify-lock.json` pinned `@gjsify/cli: 0.14.0` — so the bootstrapped CLI lost the PATH lookup to a 14-versions-stale one. Verified in a local checkout: `node_modules/.bin/gjsify -> node_modules/@gjsify/cli/lib/index.js`, version `0.14.0`.

0.14.0 predates the fix for spawning oxlint via `process.execPath` (gjsify `78ae67790`, 2026-07-03, an ancestor of **both** v0.27.1 and v0.28.0). Under GJS that path is the CLI bundle itself — which carries `#!/usr/bin/env -S gjs -m` and is executable — so the CLI re-executed **itself** with oxlint's argv. That is the failing run's output verbatim:

```
Running on GJS 1.86.0 (SpiderMonkey)
script "check" exited with code 1
```

0.14.0's own source says so, unminified — `lib/utils/oxc-resolve.js:263`: *"oxlint: always spawned via its Node ESM launcher with the current Node executable (`process.execPath`)"*.

## 2. `json-glib` was missing from the container

`test-e2e` failed with `gjsify build: no usable bundler engine under GJS — @gjsify/rolldown-native is not loadable`, and its log lists five detected native prebuilds with **rolldown-native absent from them**.

Measured on `fedora:43` with this workflow's exact package list: the engine's prebuild needs `libjson-glib-1.0.so.0`, which nothing here installed, so GIRepository cannot open its backing library. Adding `json-glib` alone makes `gjsify build --app gjs` succeed on **gjs 1.86.0**.

Declared upstream too (gjsify#994), so the installer names the missing library instead of failing with `Unsupported type void, deriving from fundamental void` — which is what it does today, and which names nothing.

## `fedora:43` stays

GJS 1.86 was suspected and **cleared by measurement**: with json-glib present it installs, constructs `BundlerSession` and builds. It is also the Fedora 43 / GNOME 49 / `org.gnome.Platform//49` baseline, so dropping it would cost more than it buys.

## About the lockfile

This is a **real upgrade**, not a format migration: 94 packages move `0.14.0 -> 0.29.0`, and the tree grows 878 → 931 entries.

Regenerated with the **published 0.29.0** — the version this CI bootstraps — for two reasons:

- gjsify writes `lockfileVersion 4` as of that release; 0.28.0 cannot read it, so a v4 file regenerated with a newer CLI would break the bootstrapped one.
- With 0.28.0 the regeneration is not even possible: it dies on `EBADPLATFORM` for `@parcel/rust-darwin-x64`, the both-block optional-dependency bug fixed in gjsify#988 and released in 0.29.0.

## What proves it

CI. It has been red on both jobs, so a green run here is the first evidence either way — and it is the only thing that can vouch for a 15-minor-version jump across 94 packages.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,7 +47,7 @@ jobs:
       - name: Install container prerequisites
         run: |
 
-          dnf install -y git tar xz findutils curl gjs libsoup3
+          dnf install -y git tar xz findutils curl gjs libsoup3 json-glib
 
           git config --global --add safe.directory "*"
       - name: Set up Node.js 24
@@ -80,7 +80,7 @@ jobs:
     steps:
       - name: Install container prerequisites
         run: |
-          dnf install -y git tar xz findutils curl gjs libsoup3
+          dnf install -y git tar xz findutils curl gjs libsoup3 json-glib
           git config --global --add safe.directory "*"
       - name: Set up Node.js 24
         uses: actions/setup-node@v4
@@ -117,7 +117,7 @@ jobs:
     steps:
       - name: Install container prerequisites
         run: |
-          dnf install -y git tar xz findutils curl gjs libsoup3
+          dnf install -y git tar xz findutils curl gjs libsoup3 json-glib
           git config --global --add safe.directory "*"
       - name: Set up Node.js 24
         uses: actions/setup-node@v4
@@ -174,7 +174,7 @@ jobs:
       - name: Install container prerequisites
         run: |
 
-          dnf install -y git tar xz findutils curl gjs libsoup3
+          dnf install -y git tar xz findutils curl gjs libsoup3 json-glib
 
           git config --global --add safe.directory "*"
       - name: Set up Node.js 24
--- a/gjsify-lock.json
+++ b/gjsify-lock.json
@@ -1,8 +1,8 @@
 {
-  "lockfileVersion": 2,
+  "lockfileVersion": 4,
   "requested": [
     "@biomejs/biome@^2.5.1",
-    "@gjsify/cli@^0.14.0",
+    "@gjsify/cli@^0.29.0",
     "@release-it/bumper@^7.0.6",
     "@release-it/conventional-changelog@^11.0.1",
     "concurrently@^10.0.3",
@@ -78,6 +78,16 @@
       "version": "2.5.1",
       "resolved": "https://registry.npmjs.org/@biomejs/biome/-/biome-2.5.1.tgz",
       "integrity": "sha512-IXWLCxKmae+rI7LOHS1B3EbVisQ6GRAWbhN9msa6KjNCyFWrvKZWR4oUdinaNssrV852OrSHuSPa95h1GPJc7Q==",
+      "optionalDependencies": {
+        "@biomejs/cli-linux-x64": "2.5.1",
+        "@biomejs/cli-win32-x64": "2.5.1",
+        "@biomejs/cli-darwin-x64": "2.5.1",
+        "@biomejs/cli-linux-arm64": "2.5.1",
+        "@biomejs/cli-win32-arm64": "2.5.1",
+        "@biomejs/cli-darwin-arm64": "2.5.1",
+        "@biomejs/cli-linux-x64-musl": "2.5.1",
+        "@biomejs/cli-linux-arm64-musl": "2.5.1"
+      },
       "bin": {
         "biome": "bin/biome"
       }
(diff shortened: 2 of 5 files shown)


JumpLink:
Pushed `51b03043` — the remaining `test-e2e` red had a **third** cause, independent of the two in the description.

## It is not json-glib

The log names it directly:

```
@gjsify/rolldown-native is NOT INSTALLED — no node_modules/@gjsify/rolldown-native
was found walking up from /__w/ts-for-gir/ts-for-gir/packages/cli
```

That is the *not-installed* branch of the diagnostic, not the *not-loadable* one. The json-glib fix in this PR is correct and stays — it just addresses a different failure than the one now showing. Corroborating detail from the same log: `Detected 5 @gjsify/* package(s) with native prebuilds`, with rolldown-native absent from the five.

## Why it was missing

```
$ npm view @gjsify/cli@0.29.0 peerDependencies peerDependenciesMeta
peerDependencies:     { "@gjsify/rolldown-native": "^0.29.0", … }
peerDependenciesMeta: { "@gjsify/rolldown-native": { "optional": true }, … }
```

An **optional** peer, which npm does not install. That is right for `@gjsify/cli`: a Node-target consumer uses the npm `rolldown` crate and has no use for a native GJS engine, so making it mandatory would push a native package onto every Node-only consumer. This repo is simply not that consumer — `build:app:gjs` cannot run without it.

So the fix belongs here and in the manifest, not in an install line: a real build dependency that was undeclared.

## Measured, before and after

| | before | after |
|---|---|---|
| `node_modules/@gjsify/rolldown-native` | absent | present |
| detected native prebuild packages | 5 | **6** (`@gjsify/rolldown-native-linux-x64` added) |
| `gjsify run build:app` | — | **exit 0**, 0 errors |
| `gjsify run build:app:gjs` | `script "build" exited with code 1` | **exit 0**, 0 errors |

Reproduced locally first: the worktree's `node_modules` carried no rolldown-native either, matching CI exactly. Lockfile 931 → 935 entries; the platform filter did its job — only `linux-x64` of the five targets was fetched.

## Ordering note

With this green, the chain the ADR depends on is unblocked: **#437 → #435 → #436**. #435 is the prerequisite for trusting a green run on #436, since `build:types` runs through `bin/ts-for-gir-dev`, which discarded its child's exit code — so the step generating ~700 `@girs/*` packages could not fail. #436 is decision 3 of gjsify's new ADR 0019 (gjsify#1006).

JumpLink:
## Retargeted to 0.30.0 — and the remaining red is now measured, not guessed

The `test-e2e` leg was still red after `51b03043` declared the engine, so the engine was not the whole story. The second cause is a **version choice**, and 0.29.0 could not have worked.

`test:cli-tarball-shape` shells out to the production packer — `tests/e2e/cli-tarball-shape/run.mjs:60-64`, `gjsify pack packages/cli --pack-destination … --json` — under **two independent 5-minute ceilings**: `execFileSync`'s own `timeout: 5 * 60 * 1000` and `--test-timeout=300000` (`tests/e2e/package.json:9`). And `@ts-for-gir/cli` declares `prepack`.

That is exactly the trigger in gjsify#1010 — *"`gjsify pack` never exits under GJS when the package declares a lifecycle script"* — measured there at **5m30s** of parking after the work had finished. Which is why the observed failure is `Error: spawnSync gjsify ETIMEDOUT` and not an assertion.

gjsify#1011 fixed it, and it landed **between the two tags**:

| | |
|---|---|
| `git merge-base --is-ancestor fb881d0db v0.29.0` | **false** |
| `git merge-base --is-ancestor fb881d0db v0.30.0` | **true** |
| v0.29.0 tagged | 2026-08-05 16:10:28 +0000 |
| #1011 merged | 2026-08-05 20:58:17 +0000 |
| v0.30.0 tagged | 2026-08-05 21:42:33 +0000 |

## Verified locally on this branch, not inferred

```
▶ @ts-for-gir/cli tarball shape (gjsify pack — production packer)
  ✔ tarball was produced (454.2653ms)
  … 8 rows …
✔ 8 pass, 0 fail — duration_ms 6434.560859
```

**6.4 seconds for the whole suite**, ~450 ms per `gjsify pack`, where 0.29.0 spent 5m30s and then died on the ceiling.

## What else changed

- Lockfile regenerated with the **published 0.30.0** — the version this CI bootstraps from `releases/latest`. Forward-reading a `lockfileVersion 4` file with an older CLI is not possible (`install-backend-native.ts:214`, `:234` — `READABLE_LOCKFILE_VERSIONS = {2, 3, 4}`), so the regenerating CLI must not be newer than the bootstrapped one. Result: 444 `0.30.0` entries, zero `0.29.0`.
- `packages/cli/templates/types-gjsify/package.json` carried a stale `@gjsify/cli: ^0.4.23`. `syncGjsifyDepsFromCli()` (`packages/cli/scripts/process-templates.mjs:131-146`) overwrites it from `packages/cli` at build time, so the value was already inert — but a stale co-existing pin is the incident that mechanism exists to prevent, so it should not read as current.
- `@gjsify/rolldown-native` stays declared. It is item 3 of gjsify#1005 and remains correct regardless of how the core-side gap is closed: post-install now reports **6** native prebuild packages, and the platform filter fetched only `linux-x64`.
- `json-glib` in the workflow stays. gjsify#994 makes the installer *name* the missing library; it does not install it.

## Deliberately not in this PR

`examples/gtk-4-template-blueprint-vite/package.json:17` still pins `@gjsify/vite-plugin-blueprint: ^0.14.0`. Touching it would flip the `examples/**` path filter on (`ci.yml:208-252`) and pull a 16-minor plugin bump into the PR that is the critical path for #435 → #436. Same stale-pin class, its own change.

PR #439, opened 2026-08-26
fix(lib): correct the nick derivation and document library use

JumpLink:
Two things, both from the same question: gjsify asked what else of its GIR-handling code is really ts-for-gir's. The audit itself lives in gjsify; the parts that belong to **this** repo are here.

## 1. The 97-vs-889 contradiction was two derivations, not two answers

gjsify's `gir.mts` said 97 members contradict a derived nick; ts-for-gir said 889. Same corpus — gjsify's generator literally reads this repo's `girs/`. Measured over all 705 files, 59 789 `<member>` elements, 40 940 with `glib:nick`:

| derivation | disagreements |
|---|---:|
| `name.replace(/_/g, "-")` | **97** |
| `name.toLowerCase().replace(/_/g, "-")` | **889** |

Neither number was wrong. They differ by exactly the **792** members whose GIR name carries uppercase and whose nick preserves it — those disagree only with a derivation that lowercases. `97 + 792 = 889`, and each repo had quoted the number for *its own* fallback. What was wrong is that two comments here quoted the 889 while naming the underscore substitution, which measures 97.

### And the lowercasing itself is a defect

Two invariants hold across the corpus, with zero exceptions:

1. **A nick differs from its member name only in which underscores became dashes** — 40 940 of 40 940. Never case, never spelling. All 836 uppercase-named members carry their name verbatim as the nick.
2. **No member lacking `glib:nick` has an uppercase name.** The fallback's entire input set is already lowercase.

(1) makes the `toLowerCase()` wrong in principle; (2) is why nobody noticed — it cannot change one byte of generated output, which is what makes dropping it safe rather than a regeneration risk.

Also worth keeping: 53 of the 97 are all-lowercase names where only *some* underscores became dashes — `audio_unit_client_activated` registers as `audio_unit-client-activated`. No mechanical derivation reaches that. That, not a count, is why the attribute wins.

### The mechanism, not just the fix

`scripts/check-nick-derivation.mjs` (wired in as `check:girs`, ~1.7 s over the corpus) prints both counts and **asserts both invariants**, so the numbers live where the corpus does instead of drifting in comments in two repositories. Verified to go red, not just green: a fixture whose nick differs beyond `_`/`-`, and one with an uppercase name and no nick, each exit 1 naming the member.

The comments now state what the numbers establish, per the convention gjsify#1328 set.

## 2. How to use ts-for-gir as a library

`packages/lib/README.md` listed features and never said how to use any of them. It now answers the question people actually arrive with — *how do I get TSX or Vue types for GTK widgets out of GIR* — and routes two of three askers away from this package:

1. **Published namespace, generate nothing.** `@girs/<ns>/surface` is the whole input a framework binding needs. Documents every export, and the four decisions it deliberately does not make: tag spelling, signal prop names, the framework's shape, property-key case. Together they are one mapped type.
2. **Private or vendored GIR, use the CLI as a subprocess.** No library API, no TypeScript build on the consumer's side.
3. **Changing what gets emitted, use the library.** With the constraint stated plainly: `exports["."]` is `./src/index.ts`, so a consumer must compile or bundle it, and must not put it in a published package's `dependencies`.

The tag-spelling example is deliberately *not* clever. The obvious camel-to-kebab conditional type resolves `GtkGLArea` to `gtk-g-l-area` — checked against `tsc`, not assumed — which is the honest reason a tag map gets hand-corrected, and why Volar's camelize round trip constrains the choice for Vue consumers.

## 3. One claim withdrawn

The header emitted into **every** generated surface said a JSX namespace "is a global declaration that a second library cannot declare without colliding". The first half is false, and gjsify#1337 corrects it upstream: the shape that collides is the `declare global` augment, and gjsify's own gtk-host ships a Solid and a React module-scoped namespace in one package. The same correction is applied to `AGENTS.md`. The observation about global augments is preserved; only its generalisation is withdrawn.

## Checks

`check:girs` ✅ · `gjsify lint` ✅ exit 0 · `check` on `@ts-for-gir/lib` + `@ts-for-gir/generator-typescript` ✅ · `@ts-for-gir-test/widget-surface` ✅ including its three controls (flag-off clean, broken fixture rejected with exit 1, halves compared).

Generated output is unchanged by the nick fix, by invariant (2) above. The surface header text does change, and the regenerated fixture confirms it.


JumpLink:
Counterpart: gjsify/gjsify#1340 — carries the audit this came out of (what else could move here, and the line that decides it), plus the `tier.mjs` rule that now holds ADR 0019's build-step-free boundary.

PR #445, opened 2026-08-28
chore(deps): update every external dependency

JumpLink:
Stacked on #444. Merge that one first.

25 external dependencies to latest, six of them majors, plus a full lockfile re-resolve. Generated `@girs/*` workspaces are excluded, since their manifests come from a template and a bump there is churn that the next `build:types` overwrites.

## The majors, and what I checked

release-it 20 to 21, `@release-it/bumper` 7 to 8, `@release-it/conventional-changelog` 11 to 12, `conventional-changelog-conventionalcommits` 9 to 10.

This is the quartet that broke v4.1.0. A conventionalcommits major resolved to an API-incompatible template set, and `beforeRelease` wrote a version heading with no commit list under it. `scripts/check-release-changelog.mjs` exists because of that release.

So I ran it rather than trusting the version numbers. A dry run on this branch produces a complete changelog:

```
## [4.3.0-rc.0](.../compare/v4.2.0...v4.3.0-rc.0) (2026-08-28)
### Features
* **girs:** update GIRs to GNOME 51 beta (8fda5b2)
* **lib:** carry the Mutter patches into GNOME 51 (d083879)
...
### Maintenance
* **deps:** update every external dependency (a8a0d74)
```

All sections populated, every commit listed. The v4.1.0 failure mode does not recur.

`@types/node` 25 to 26 and cosmiconfig 9 to 10 needed no source change.

`@gjsify/cli` and `@gjsify/rolldown-native` go 0.30.0 to 0.44.0 together. They have to move in lockstep. The CLI's bundler engine under GJS is rolldown-native, it is not a dependency of the CLI, and a globally installed copy hides a stale pin until someone builds from a clean tree, which is what CI always does.

## What oxlint found

oxlint 1.69 to 1.80 turned up one real finding. `no-irregular-whitespace` flags the zero-width space in `surface/model.ts`'s doc comment, which was there so the comment could contain a comment terminator.

The doc comment now says it in words. The space stays where it does actual work, inside the `replace()` that breaks a terminator coming out of a GIR doc, and carries a note explaining itself, because the next reader cannot see it either.

## TypeScript 7 is out, and why

typedoc 0.28.20 peer-depends on `5.0.x || ... || 5.9.x || 6.0.x`. 7.0.2 is not installable while typedoc is a direct dependency.

Bumping the declaration anyway did not fail. It resolved to 5.9.3 and said nothing, which is worse than an error: a version nobody asked for, chosen quietly.

That led to something separate, which predates this branch. The repo declares `typescript ^6.0.3` and installs **5.9.3**, before and after `--refresh-lockfile`. The lockfile is committed, so CI has been type-checking on 5.9.3 as well. Worth chasing on its own; nothing here causes it.

## Validation

```
check          0   (lint, app, girs, pm, prereqs, docs)
build:app      0
test:tests     0
release-it     dry run produces a full changelog
```

https://claude.ai/code/session_016ci5gmUojZSbm5JXkgwYKj


JumpLink:
Added one more commit here, `ci: run on PRs whose base branch has a slash`.

While waiting for #444 I noticed this PR and #446 had only PR Lint and no CI at all. The cause is in `ci.yml`:

```yaml
pull_request:
  branches:
    - "*"
```

That filters by **base** branch, and a GitHub filter `*` does not cross a slash. A PR based on `feat/gnome-51-beta` or `chore/deps-latest` matched nothing, so the workflow never ran for it.

It fails quietly, which is the annoying part. The checks are absent rather than red, and on the PR page that reads as "still queued". Stacked PRs are where it happens.

Measured on the three open PRs at once: `pr-lint.yml` declares no branches filter and ran on all three, `ci.yml` ran only on #444, whose base is `main`. After changing `"*"` to `"**"` and pushing, CI queued for this branch and for #446 within seconds.

The commit sits here rather than in #444 so it takes effect immediately without restarting that PR's 30-minute `build-validate`.

JumpLink:
GitHub closed this when #444 merged and `feat/gnome-51-beta` was deleted, and a closed PR cannot be reopened or retargeted. Continued in #447 with the same three commits, rebased onto `main`.

PR #450, opened 2026-08-29
feat(gir-files): publish the original GIR XML files

JumpLink:
## What this adds

`@ts-for-gir/gir-files` — one pooled package carrying the original **GIR XML** that
`@girs/*` is generated from, with a per-namespace provenance and licence record.

```ts
import { readGirXml, getGirFile, listGirFiles, extractGirFiles } from "@ts-for-gir/gir-files";

const xml  = readGirXml("Gtk-4.0");   // the full <repository> document
const info = getGirFile("Gtk-4.0");   // license, sourcePackage, sourceRpm, sizes
extractGirFiles("./girs");            // plain .gir files, for --girDirectories consumers
```

## Why

GIR XML and the compiled typelib carry **different** data, and the generated `.d.ts`
carries neither in full:

| | GIR XML | typelib | generated `.d.ts` |
|---|---|---|---|
| enum nicks (`glib:nick`) | yes | yes | in `@girs/<ns>/surface` |
| **documentation prose (`<doc>`)** | **yes** | no | no |
| deprecation reasons (`<doc-deprecated>`) | yes | no | no |

`Gtk-4.0.gir` holds **18,386 `<doc>` elements**. A tool that wants to explain *why* — a
property ledger with reasons and doc text, a conformance checker, a docs builder — has to
read the XML, and until now the only way to get it was to install the distro `-devel`
packages and hope their versions matched the types being compiled against.

## Scoping against ADR 0019 § 2

gjsify's [ADR 0019 § 2](https://github.com/gjsify/gjsify/blob/main/docs/adr/0019-ts-for-gir-as-library.md)
decided *"the `.gir` travels with the RUNTIME package, never with the type package"*, and
rejects shipping `.gir` inside `@girs/*` because **"a `.gir` in a type package proves
nothing about the library the host will load"**.

That argument is correct and this package does not contradict it. What § 2 rejects is a
`.gir` used to answer *"what will the host load?"* — the drift it was written about, where
`tsc` exits 0 and the runtime throws `TypeError`. This package answers a different
question: **what did the corpus that generated `@girs/*` at this version say?** That is a
build-time correspondence between input and output, not a claim about any host.

Three things keep it on the right side of the line:

- It is neither a type package nor a runtime package, so it adds no `.gir` to any `@girs/*`
  tarball and duplicates nothing across ~705 packages — the ADR's other objection.
- Every entry names its own origin (`sourcePackage: "gtk4-devel"`,
  `sourceRpm: "gtk4-4.23.3-1.fc45.src.rpm"`), so a consumer can see it is reading Fedora's
  package and not the GTK on this machine.
- The README says outright that it must not be used to decide what is available at
  runtime, and points at `@gjsify/gtk-runtime-*` for that.

ADR 0019 § 5 draws the same boundary in this package's favour: *"ts-for-gir knows GIR as
XML — parsed headlessly, in CI, with no GTK installed and no typelib loaded."* The
consumers here are generators, conformance checkers and documentation builders, which never
load the library at all. § 4 also states explicitly that the boundary *"does not restrict
where GIR-derived DATA may flow"*.

**Venue**: `gjsify/types` has issues disabled (`has_issues: false`) and is generator output;
`ts-for-gir` owns `girs/` and the release train, so it is the right repo. No existing issue
or PR covers this.

## Licensing — please read this part

**`girs/` is not our work.** Every file is a verbatim copy of a Fedora RPM's contents
(`scripts/fetch-fedora-girs.sh` harvests them with `dnf repoquery -f '*.gir'`), a
`g-ir-generate` decompile, or a gnome-shell build artefact. **Not one of the 718 files
carries a licence header** — `grep -lE 'SPDX-License|Copyright \(c\)' girs/*.gir` matches
zero files — so the terms are inherited from the project each was generated from.

Before this PR the repository recorded **nothing** about that: no NOTICE, no carve-out in
`LICENSE`, and `git log -i --grep=licen --oneline origin/main` over the whole history returns zero commits. The
generated `@girs/*` packages stamp a blanket `"license": "MIT"`
(`packages/templates/templates/package.json:107`) regardless of the source namespace.

The terms are real copyleft. Over the 510 attributed files, **93 distinct licence
expressions**, the top of the distribution being:

| expression | files |
|---|---|
| `LGPL-2.1-or-later` | 114 |
| `GPL-2.0-or-later` | 103 |
| `LGPL-2.0-or-later` | 46 |
| `LicenseRef-Callaway-LGPLv2+` | 30 |
| `LGPL-3.0-or-later` | 13 |
| `GPL-3.0-or-later` | 10 |

So this PR builds the mechanism that makes attribution possible rather than assuming an
answer:

- **`scripts/build-gir-provenance.sh`** queries Fedora's repository metadata for the
  `%{license}`, `%{url}` and `%{sourcerpm}` of the package owning each `.gir`. It installs
  nothing — metadata only. Namespaces Fedora cannot place fall back to this repo's own
  curated `@ts-for-gir/gir-module-metadata` records, which cover gnome-shell (built from
  source, in no RPM) and, via `versionAgnostic`, the historical namespace versions.
- **A file ships only if the manifest attributes it.** 510 of 718 today; the 208 held back
  are listed in `provenance.json` → `unattributed` (108 are older versions of namespaces
  that *are* attributed, 100 are namespaces current Fedora no longer ships). `readGirXml()`
  on a withheld namespace throws and says why.
- **`NOTICE.md`** is generated from the manifest and lists every shipped file with its
  licence and origin. `package.json` says `"license": "SEE LICENSE IN NOTICE.md"` — no
  single identifier is true of these contents, and claiming MIT would be false.

### What still needs a human, before this is published

**I did not, and could not, make the licensing determination — only the evidence for one.**
Two limits are stated in the code and repeated here:

1. A distribution's `License:` tag is an expression over **everything in that RPM**, not a
   per-file finding. `gtk4-devel` declares ten licences AND-ed together; which of them
   governs `Gtk-4.0.gir` specifically is not something the tag answers.
2. Some values are Fedora-internal legacy identifiers (`LicenseRef-Callaway-LGPLv2+`,
   30 files) rather than real SPDX, and will not resolve in an SPDX tool.

LGPL and GPL do permit verbatim redistribution of source with notices preserved, which is
what distros and `gtk-rs/gir-files` rely on, and the NOTICE is built to satisfy that. But
**the decision to publish is yours, not mine.** Nothing here has been published; the
package would ship only on a future release cut, and it would be reasonable to hold it
until that review happens. There is also a separate, pre-existing question this surfaces
but does not touch: whether the blanket `MIT` on the generated `@girs/*` packages is right.

## Shape and size

One pooled package, not one per namespace. Per-namespace would mean **718 new npm names**;
in `packages/*` — the home chosen below — each would need its own manual first-publish plus
Trusted Publisher bootstrap, since that train publishes over OIDC. (To be precise: a
per-namespace set placed in `gjsify/types` instead would avoid the bootstrap, because that
train authenticates with a token and its publisher treats a 404 as "new package". It would
still be 718 names to version and 718 tarballs, and it would inherit that repo's problem
that nothing bumps versions there.) And GIR consumers are tools — generators, conformance
checkers, docs builders — which usually want many namespaces at once.

The payload ships **gzipped per file**. Measured over the 510 shipped namespaces via
`npm pack --dry-run --json`:

| shape | tarball | installed |
|---|---|---|
| plain `.gir` | 27.5 MB | 308.0 MB |
| per-file `.gir.gz` (shipped) | **27.6 MB** | **27.9 MB** |

The tarball costs 0.1 MB (+0.4 %) and the installed footprint is **11.0× smaller**. This
matters: 308 MB unpacked would be ~2.5× `aws-cdk-lib` (124.3 MB), one of the largest
widely-installed packages, and npmjs.com publishes no clear hard limit to bet against.
27.9 MB is ordinary — `@swc/core-linux-x64-gnu` is 30.5 MB.

**Do not read the compression ratio as files sharing structure.** They do not: gzip's 32 KB
window captures almost nothing across a 376 MB set. Over the full pool, one `tar.gz` of all
718 files is **32.4 MB** while gzipping each separately and summing is **32.7 MB** — 0.9 %
apart. Pooling into one stream buys essentially nothing, which is exactly *why*
pre-compressing per file gives up essentially nothing. (`zstd --long` or `xz` would do
considerably better here, but npm tarballs are gzip.)

## Versioning

The package lives in `packages/*`, so `@release-it/bumper` versions it with the train and
`@ts-for-gir/gir-files@X` carries the XML that produced `@girs/*@X`. No hand-maintained
range can drift out of step — the failure that has reddened installs on this project twice.

The cost of that home is that `release-app.yml` publishes via OIDC, where a new name needs
a bootstrap. Hence the third commit.

## Release-path fix included

`publish:app` passed `--trusted` without `--tolerate-untrusted-new`. Trusted Publishing
requires the package to already exist, so the **first** release carrying any new
`@ts-for-gir/*` name would 404 on the OIDC exchange, hard-fail, and take every
alphabetically later package down with it — gjsify's docs record exactly this as the
v0.4.20 incident (60+ packages stuck a version behind). The flag is the documented remedy
and pairs with the `--tolerate-republish` already present.

**Before the release that ships this**, a maintainer still has to do the manual bootstrap:

```bash
gjsify workspace @ts-for-gir/gir-files build
cd packages/gir-files && gjsify publish --access public --otp <code>
gjsify trust @ts-for-gir/gir-files --workflow release-app.yml   # NOT the default release.yml
```

## The gate, watched red first

`tests/gir-files` (Vitest, picked up automatically by `test:tests`). The failure it exists
for is a tarball with **zero `.gir` files at exit 0** — nothing else in the repo would
notice, since there is no type to compile and no import to resolve.

**Red 1 — payload built empty** (builder skipped / empty manifest):

```
REAL vitest exit=1
     × is not empty 3ms
     × contains exactly the namespaces the manifest attributes 6ms
     × has samples to read 0ms
 Test Files  1 failed (1)
      Tests  3 failed | 10 passed (13)
```

**Red 2 — one unattributed `.gir` dropped into the payload** (the licensing invariant):

```
injecting unattributed namespace: Ags-6.0
REAL vitest exit=1
 FAIL  src/gir-files.test.ts > licensing > ships no unattributed namespace
AssertionError: expected [ 'Ags-6.0' ] to deeply equal []
 Test Files  1 failed (1)
      Tests  2 failed | 17 passed (19)
```

**Green — payload rebuilt from scratch:**

```
REAL exit=0
==> payload: 510 files
    raw  308.0 MB -> gzip 27.6 MB (11.2x)
 Test Files  1 passed (1)
      Tests  19 passed (19)
```

Every count in the gate is **derived from the manifest**, never written down: the pool
grows (GNOME 51 added 13 namespaces in one commit — 705 → 718), so a hard-coded number
would go red for the wrong reason on the next refresh.

The harvester itself has the same discipline, learned the hard way — its first two runs
exited 0 having attributed nothing via Fedora, because `podman run` without `-i` gives the
container no stdin, and because dnf's `--qf` interprets `\n` but not `\t` (without the
newline all 288 packages come back as one line, which `sort -u` reports as one package). It
now refuses to write a manifest on an empty or implausibly small result.

## Verification

```
check:lint    exit=0      check:app     exit=0 (tsc over @ts-for-gir/gir-files)
check:pm      exit=0      check:prereqs exit=0
check:girs    exit=0      check:docs    exit=0
gjsify install --immutable exit=0       (lockfile unchanged; no new registry deps)
gjsify run test:tests      exit=0       (gir-files 19/19, plus the existing suites)
gjsify run build:app       exit=0       (chains the payload; tree stays clean)
```

## Reproducing every number

Measured on `origin/main` @ `7e97f3f3` (`git rev-list --count HEAD..origin/main` = 0).
Note the pool grew in `575ec30d` (GNOME 51): figures taken before it say 705 files /
361.8 MB, and `Gtk-4.0.gir` is now 8.04 MB, not 6.20 MB.

```bash
# pool size and count
find girs -name '*.gir' -printf '%s\n' | awk '{s+=$1;n++} END {printf "%d files, %.1f MB\n", n, s/1048576}'
#=> 718 files, 376.0 MB

# one tar.gz over the whole pool vs. the sum of the files gzipped separately
tar cf - --sort=name $(find girs -name '*.gir' | sort) | gzip -6 | wc -c      #=> 33987103  (32.4 MB)
find girs -name '*.gir' -print0 | xargs -0 -P8 -I{} sh -c 'gzip -6 -c "{}" | wc -c' \
  | awk '{s+=$1} END {printf "%.1f MB\n", s/1048576}'                         #=> 32.7 MB   (0.9 % apart)

# the shipped set, both shapes
scripts/build-gir-provenance.sh                       #=> 718 files, 510 attributed, 208 not, 93 licences
node packages/gir-files/scripts/build-payload.mjs     #=> 510 files, 308.0 MB -> 27.6 MB (11.2x)
cd packages/gir-files && npm pack --dry-run --json    #=> size 27.6 MB, unpackedSize 27.9 MB, 510 .gir.gz

# comparison packages
curl -s https://registry.npmjs.org/aws-cdk-lib/latest | jq '.dist.unpackedSize'   #=> 124.3 MB (vs 27.9 MB here)
curl -s https://registry.npmjs.org/@swc/core-linux-x64-gnu/latest | jq '.dist.unpackedSize'
```

## Not done

- Nothing published to npm; no bootstrap performed.
- `girs/` is untouched, and no submodule was modified.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,3 +49,8 @@ repo/
 # Staging dir of scripts/fetch-fedora-girs.sh (container harvest, reviewed then
 # copied into girs/ — never a source of truth on its own)
 .fedora-girs-*/
+
+# Generated payload of packages/gir-files (rebuilt by build:app from girs/).
+# NOTICE.md and provenance.json ARE committed: they are the licence record and
+# have to be reviewable in a diff.
+packages/gir-files/payload/
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "update:submodules": "git submodule foreach git pull",
     "update:packages": "gjsify upgrade --latest",
     "update": "gjsify run update:submodules && gjsify run update:packages",
-    "publish:app": "gjsify run build:app && gjsify foreach -v -p --no-private --include '@ts-for-gir/*' --include '@gi.ts/*' --exec -- gjsify publish --trusted --tolerate-republish --tag latest --access public --provenance",
+    "publish:app": "gjsify run build:app && gjsify foreach -v -p --no-private --include '@ts-for-gir/*' --include '@gi.ts/*' --exec -- gjsify publish --trusted --tolerate-untrusted-new --tolerate-republish --tag latest --access public --provenance",
     "publish:types": "gjsify foreach -v -p --no-private --include '@girs/*' --exec -- gjsify publish --tolerate-republish --tag latest --access public",
     "publish": "gjsify run publish:app && gjsify run publish:types",
     "test:types": "gjsify run build:types && gjsify run check:types",
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "start": "ts-for-gir-dev",
     "start:prod": "ts-for-gir",
     "format": "gjsify fix",
-    "build:app": "gjsify workspace @ts-for-gir/cli build",
+    "build:app": "gjsify workspace @ts-for-gir/cli build && gjsify workspace @ts-for-gir/gir-files build",
     "build:app:gjs": "gjsify workspace @ts-for-gir/cli build:gjs",
     "build:examples": "gjsify foreach build -v -p --include '@ts-for-gir-example/*'",
     "build:types": "ts-for-gir-dev generate --configName='.ts-for-gir.packages-all.rc.js' --girDirectories='./girs' --workspace=true --outdir='./types-dev' && gjsify install",
--- a/packages/gir-files/NOTICE.md
+++ b/packages/gir-files/NOTICE.md
@@ -1,0 +1,639 @@
+# NOTICE — third-party content in `@ts-for-gir/gir-files`
+
+This package redistributes **GObject Introspection (GIR) XML files that this project did
+not author**. They are verbatim copies of files shipped by upstream projects and by the
+Fedora packages built from them.
+
+## Licensing
+
+The packaging around them — `src/`, `scripts/`, `provenance.json` — is **Apache-2.0**,
+like the rest of ts-for-gir.
+
+The `payload/` files are **not**. Each carries the licence of the project it was generated
+from, and those are predominantly copyleft. The table below records, for every file in this
+package, the licence expression declared by the upstream source it came from. That is why
+`package.json` says `"license": "SEE LICENSE IN NOTICE.md"` rather than naming one licence:
+no single identifier is true of this package's contents.
+
+**This table is evidence, not a legal determination.** A distribution's `License:` tag is an
+expression covering everything in that package, not a per-file finding, and nothing here is
+legal advice. If you redistribute these files onward, satisfy the terms below yourself.
+
+Files that no source could attribute are **not included** in this package. See
+`provenance.json` → `unattributed` for what was left out and why.
+
+## Licence expressions present (93 distinct, over 510 files)
+
+- `LGPL-2.1-or-later` — 114 file(s)
+- `GPL-2.0-or-later` — 103 file(s)
+- `LGPL-2.0-or-later` — 46 file(s)
+- `LicenseRef-Callaway-LGPLv2+` — 30 file(s)
+- `LGPL-2.1-or-later AND LGPL-2.0-or-later AND (MIT OR LGPL-2.1-or-later) AND MPL-1.1 AND BSD-2-Clause AND BSD-3-Clause AND BSD-2-Clause-Views AND (BSD-2-Clause AND DOC) AND MIT-Festival AND (LGPL-2.0-or-later AND LicenseRef-Fedora-Public-Domain) AND (MPL-1.1 OR LGPL-2.0-or-later OR MIT) AND BSD-3-Clause WITH AdditionRef-Dart AND MIT AND GPL-2.0-only WITH Linux-syscall-note` — 19 file(s)
+- `GPL-2.0-or-later AND LGPL-2.0-or-later AND LGPL-2.1-or-later AND BSD-2-Clause` — 14 file(s)
+- `LGPL-3.0-or-later` — 13 file(s)
+- `GPL-3.0-or-later` — 10 file(s)
+- `MIT` — 7 file(s)
+- `GPL-2.0-or-later AND LGPL-2.1-or-later` — 6 file(s)
+- `LGPL-2.1-only AND BSD-2-Clause AND BSD-3-Clause AND ISC AND bzip2-1.0.6 AND NCSA AND MIT AND GPL-2.0-only AND MPL-1.1 AND SunPro AND Apache-2.0 AND GPL-3.0-or-later WITH Bison-exception-2.2 AND MPL-2.0 AND OFL-1.1 AND (AFL-2.0 OR GPL-2.0-or-later) AND BSD-Source-Code AND BSD-2-Clause-Views AND LGPL-2.1-or-later AND (NCSA OR MIT) AND Apache-2.0 WITH LLVM-exception AND BSL-1.0` — 6 file(s)
+- `GPL-2.0-or-later AND LicenseRef-Callaway-LGPLv2+` — 5 file(s)
+- `LGPL-2.0-or-later AND LGPL-2.1-or-later AND Apache-2.0 AND CC0-1.0 AND MIT AND MIT-open-group AND HPND-sell-variant AND GPL-2.0-or-later AND GPL-3.0-or-later AND OFL-1.1` — 5 file(s)
+- `Apache-2.0` — 4 file(s)
+- `LGPL-2.1-only AND GPL-2.0-only AND GPL-2.0-or-later` — 4 file(s)
+- `GPL-3.0-or-later AND LGPL-3.0-or-later` — 4 file(s)
+- `LGPL-2.1-or-later AND CC-BY-SA-3.0` — 4 file(s)
+- `LicenseRef-Callaway-LGPLv2` — 3 file(s)
+- `GPL-3.0-or-later AND AGPL-3.0-or-later AND LicenseRef-Callaway-GFDL` — 3 file(s)
+- `GPL-2.0-or-later AND LGPL-2.0-or-later AND LGPL-2.1-or-later AND LGPL-3.0-or-later` — 3 file(s)
+- `(LGPL-3.0-only OR LGPL-2.1-only) AND GPL-3.0-only` — 3 file(s)
+- `LGPL-2.1-or-later AND Apache-2.0` — 3 file(s)
+- `LGPL-2.1-or-later AND LicenseRef-Fedora-Public-Domain AND FSFULLRWD AND (LGPL-3.0-or-later OR CC-BY-SA-3.0) AND (MPL-1.1 OR GPL-2.0-or-later OR LGPL-2.1-or-later) AND GCR-docs` — 3 file(s)
+- `LGPL-2.1-only` — 3 file(s)
+- `GPL-3.0-only` — 3 file(s)
+- `LGPL-3.0-only AND LGPL-2.1-or-later` — 3 file(s)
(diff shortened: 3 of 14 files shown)


JumpLink (commented):
The package is worth having and the licensing work is the strongest part of it. One load-bearing premise does not hold, and it is stated in four places — including a test comment, which is where a wrong reason does the most damage later.

## The documentation premise is measurably false

> | **documentation prose (`<doc>`)** | **yes** | no | **no** |

The generated `.d.ts` **does** carry the prose. Measured on `types-dev/gtk-4.0/gtk-4.0.d.ts` at `main`:

```
grep -c "^\s*\*"        gtk-4.0.d.ts   # 87870 comment lines
grep -c "@param"        gtk-4.0.d.ts   #  7983
grep -c "@returns"      gtk-4.0.d.ts   #  2937
```

It reaches methods, properties (with `@default` and `@construct-only`), enum members and signals alike:

```ts
/**
 * The accessible role used by the AT context.
 *
 * Depending on the given role, different states and properties can be
 * set or retrieved.
 * @default Gtk.AccessibleRole.NONE
 */
get accessible_role(): AccessibleRole;
```

The **deprecation** row was correct when written — 295 `@deprecated` against 780 `<doc-deprecated>` — but for a different reason than the table gives: the metadata was parsed correctly and then dropped during the rebuild into a class member. #453 fixes that at the source; on Gtk-4.0, `@deprecated` goes 295 → 1669 and `@since` 467 → 2779. So after #453 the `.d.ts` carries the reasons too, and the row is false in both halves.

**Where it is stated**, so none of these outlive the correction:

- the table in this PR description
- `packages/gir-files/src/index.ts:25-26` — *"Neither the generated `.d.ts` nor the typelib carries the `<doc>` prose"*
- `packages/gir-files/README.md:30,36`
- `tests/gir-files/src/gir-files.test.ts:139-140` — *"The reason this package exists: `<doc>` survives only in the XML."* The assertion itself is fine (the XML does carry prose); only its stated reason is wrong, which is the version that gets quoted back later.

## The argument that does hold is already in the PR, just not as the headline

Two things the XML has that TSDoc will never carry, and they are both about machines rather than readers:

- **Structure that is not prose.** In `Gtk-4.0.gir` alone: 15,432 `transfer-ownership` attributes, 807 `introspectable="0"` (i.e. exactly what is deliberately absent from the types), plus array-length parameters, C identifiers and the original nullability.
- **Parseability in CI.** A gate cannot meaningfully parse `.d.ts` comments to answer *"did the GIR say this?"*.

That second one has concrete evidence from today rather than being hypothetical. Working on #452 I needed a raw XML scan three times — as an independent third reader confirming the four child holders, for the 17-vs-4 measurement that caught a rule matching `St.Bin` and eleven `Mx.*` through method names alone, and for the doc-coverage measurement above. Every one of them ran against a local ts-for-gir checkout, which is precisely the thing gjsify's `status/open-todos.md` records as *not something CI can reproduce*. This package is what turns those into ordinary gate checks.

Suggested replacement for the table's framing: keep the enum-nick row, drop the two documentation rows, and add `transfer-ownership` / `introspectable="0"` / array-length — i.e. compare *machine-readable structure*, not prose.

## The merge decision is the licensing one, not this

Worth saying plainly because the correction above is cosmetic by comparison. The header claim verifies:

```
grep -lE 'SPDX-License|Copyright \(c\)' girs/*.gir | wc -l   # 0
grep -liE 'licen[sc]e'                  girs/*.gir | wc -l   # 65   <- the control: the grep works
```

So: 510 files, 93 licence expressions, mostly LGPL/GPL copyleft, no per-file determination possible from a distro `License:` tag, and 30 files carrying Fedora-internal identifiers that will not resolve in an SPDX tool. The PR is right that this needs a human decision and right not to make it. That is the gate on merging, and the `NOTICE.md` mechanism is the right shape for it either way.

One thing the PR raises and leaves open that deserves its own issue rather than being lost here: whether the blanket `"license": "MIT"` on every generated `@girs/*` package is correct, given the same sources.

PR #461, opened 2026-09-01
fix(lib): lint the core package, and fix the hang it hid

JumpLink:
## What

`.oxlintrc.json`'s `ignorePatterns` carried `"**/lib"`, meant for the `lib/` build output that nine `packages/*/tsconfig.json` files declare as their `outDir`. As written it also matched `packages/lib` — the source package holding the whole GIR model, which every other package is built on. **It had never been linted.**

Narrowed to `packages/*/lib`, which keeps the intent and exposes the source.

## The blindness was real

Measured tree-wide, with a control in the row:

| probe | `ignorePatterns` | exit |
|---|---|---|
| `debugger;` in `packages/lib/src/gir/signal.ts` | `**/lib` | **0** — not seen |
| the same statement | `packages/*/lib` | **1** — reported |
| `debugger;` in `packages/generator-typescript/src/generators/signal-generator.ts` | `**/lib` | **1** — seen |

Naming the file explicitly *does* report it — only the tree-wide run applies `ignorePatterns` — so a per-file check is not evidence here.

Unblinding surfaced **20 errors and 2 warnings**, all in `packages/lib`.

## And what it hid was not clean

`no-unreachable` flagged `return "ThisShouldNeverHappen"` after a `while (true)` in `getGenericNames`. The dead code was a tombstone for a **live defect**: the generator produced its first 25 names and then spun forever without yielding.

The yield was gated on `i === startIteration`, true only on the first pass, so once the loop incremented the counter nothing came out of it again. The 26th call to `createGenericNameGenerator()`'s returned function never returns — an infinite loop, not a wrong value. `createGenericNameGeneratorAt` lowers the ceiling further, because a copy resumes at the last name the original issued: a class copied at `"Z"` had exactly one name left before the same hang. Reproduced in-process; the caller does not come back.

**Generated types cannot change.** Measured over every reachable `start` (the 25 bare letters plus the default), old and new emit an identical prefix — the old one simply hangs where the new one continues.

The dead `return` existed so `next().value` typed as `string`; removing it without a replacement breaks `tsc` at both call sites (`TS2322: Type 'string | void' is not assignable to type 'string'`). A `Generator<string, never>` return type states the same thing honestly.

## The rest was style

The other 19 errors are `unicorn(no-useless-spread)`, and the 2 warnings are `typescript(consistent-type-imports)`. No behaviour change:

- **`[...xs.map(f)]` in the `copy()` methods and the base constructor.** `.map()` always allocates, so the spread copies a copy. The freshness callers depend on — no aliasing between a class and the one it was copied from — comes from `.map()`, not from the spread. The neighbouring `[...this.interfaces]` and `[...this.generics]` stay: there the spread *is* the copy, and the rule correctly leaves them alone.
- **`GirModule.accept` keeps its snapshot.** It writes each visited member back into the map it is iterating. Re-setting an existing key happens to be safe on a live `Map` iterator, but that is a subtle rule to rest on for no gain, so the array just moves into a named local.

## Guard

New `tests/generic-names` suite, picked up by `test:tests`. It pins the names actually handed out and drives the generator well past the old ceiling.

It runs the generator **from a child process under a hard timeout**. A generator that stops yielding cannot be interrupted in-process — not even by vitest's own test timeout, because `execFileSync` blocks the thread — so an in-suite guard would hang CI instead of failing it. Verified against the old generator: 4 of 6 tests fail in 117 s with an attributable message, rather than hanging.

No lockfile change: the new workspace's deps were already resolved.

## Verification

Real exit codes, all at HEAD:

| command | exit |
|---|---|
| `gjsify lint` (bare, tree-wide) | 0 |
| `gjsify run check:app` | 0 |
| `gjsify run check:girs` | 0 |
| `gjsify run check:docs` | 0 |
| `gjsify run test:tests` | 0 — 35 tests, incl. the 6 new |
| `biome format` over the touched files | 0 |

`packages/lib` now reports **0 findings**. The 10 warnings that remain tree-wide are pre-existing on `main` in `examples/`, `packages/cli`, `packages/generator-json` and `packages/typedoc-theme`, and are left alone.

Two notes on the toolchain, both pre-existing and neither caused here:

- `gjsify format --check` exits 1 with *"oxfmt not found"*. This repo formats with Biome (`biome.json`, `@biomejs/biome` devDependency) and does not install `oxfmt`; the failure reproduces identically on files this branch never touched. Biome is used as the format gate above.
- `.oxfmtrc.json` carried the same `"**/lib"` entry and gets the same narrowing. That half is unmeasured for the same reason — but Biome never excluded `lib` and reports no drift across all 143 files in `packages/lib/src`, so the change is a no-op today and disarms the trap for whoever installs `oxfmt`.

### DIFF
--- a/.oxfmtrc.json
+++ b/.oxfmtrc.json
@@ -20,7 +20,7 @@
     "**/*.gir",
     "**/node_modules",
     "**/dist",
-    "**/lib",
+    "packages/*/lib",
     "**/build",
     "**/coverage",
     "refs",
--- a/.oxlintrc.json
+++ b/.oxlintrc.json
@@ -29,7 +29,7 @@
     "**/*.cjs",
     "**/node_modules",
     "**/dist",
-    "**/lib",
+    "packages/*/lib",
     "**/build",
     "**/coverage",
     "refs",
--- a/packages/lib/src/gir-module.ts
+++ b/packages/lib/src/gir-module.ts
@@ -296,7 +296,12 @@ export class GirModule implements IGirModule {
 	}
 
 	accept(visitor: GirVisitor) {
-		for (const key of [...this.members.keys()]) {
+		// Snapshot the keys: the loop writes each visited member back into the very
+		// map it is walking, and a live iterator only survives that by a subtle rule
+		// about re-setting existing keys. Not worth depending on.
+		const keys = [...this.members.keys()];
+
+		for (const key of keys) {
 			const member = this.members.get(key);
 
 			if (!member) continue;
--- a/packages/lib/src/gir/generics.ts
+++ b/packages/lib/src/gir/generics.ts
@@ -26,31 +26,30 @@ const GenericNames = [
 	"Z",
 ];
 
-export function* getGenericNames(start: string = "A") {
-	let names = GenericNames.map((s) => `${s}`);
+/**
+ * An unbounded supply of generic parameter names: `A`…`Z`, then `A1`…`Z1`,
+ * then `A2`…`Z2`, and so on.
+ *
+ * `start` resumes the sequence at an already-issued name, so a copied class
+ * keeps handing out the names the original had not reached yet. The first
+ * pass begins at that letter; every later pass covers the full alphabet.
+ *
+ * The generator is infinite by design and therefore never returns — that is
+ * what the `never` return type states, so callers can treat `next().value`
+ * as a plain `string`.
+ */
+export function* getGenericNames(start: string = "A"): Generator<string, never, unknown> {
+	// A name is a letter plus the pass number, with pass 0 left bare ("A", not "A0").
 	const startIteration = Number.parseInt(start.slice(1) || "0", 10);
+	const startPosition = GenericNames.indexOf(start[0]);
 
-	let i = startIteration;
+	for (let iteration = startIteration; ; iteration++) {
+		for (const [position, letter] of GenericNames.entries()) {
+			if (iteration === startIteration && position < startPosition) continue;
 
-	names = names.map((s) => (i === 0 ? s : `${s}${i}`));
-
-	const StartLetter = start[0];
-	const position = GenericNames.indexOf(StartLetter);
-
-	while (true) {
-		for (const name of names) {
-			if (i === startIteration && GenericNames.indexOf(name) >= position) {
-				yield name;
-			}
+			yield iteration === 0 ? letter : `${letter}${iteration}`;
 		}
-
-		names = names.map((s) => `${s}${i}`);
-
-		i++;
 	}
-
-	// This will always be a string return.
-	return "ThisShouldNeverHappen";
 }
 
 export function createGenericNameGenerator(): () => string {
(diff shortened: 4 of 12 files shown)


JumpLink:
Independent review — findings, measurements, and one fix pushed.

## The green checks were stale

All seven checks ran at 19:39Z on Sep 1; #460 merged into main at 21:55Z. #460 touches the same `IntrospectedInterface.copy` hunk this branch rewrites, so the PR had gone `CONFLICTING` — and #460's new line is itself the pattern this branch bans:

```ts
iface.signals = [...signals.map((sig) => sig.copy({ parent: iface }))];
```

Measured on the merged tree: with that spread kept, bare tree-wide `gjsify lint` exits **1** (`unicorn(no-useless-spread)` at introspected-classes.ts:1515); with it removed, **0**. Merging this PR as it stood would have turned main red on the first post-merge lint.

Pushed `7401ad90`: a merge of `main` resolving the conflict by applying the branch's own transformation to the incoming line. Local gates on the merged tree, real exit codes: `gjsify lint` 0, `check:app` 0, `check:girs` 0, `check:pm` 0, `check:prereqs` 0, `check:docs` 0, `test:tests` 0 (all suites incl. widget-vocabulary and the 6 new).

## The hang is real — and unreachable in the shipped corpus

Reproduced out-of-process against the old generator: default start emits 25 names, the 26th `next()` never returns (killed by timeout); start `"Z"` emits 1. Old output is an exact prefix of the new generator's output for the default and **all 25 bare letters** — the full reachable start set, since the old generator could never issue a suffixed name. (Suffixed starts hang the old generator on the *first* call, but nothing could produce one.)

Reachability, measured rather than reasoned: instrumented `getGenericNames` with per-instance counters and ran full generation over the 718-GIR corpus (packages-all config). **57,595 generator instances; the busiest drew 3 names** (max per start: A→3, B→1, C→1) against the old ceiling of 25. So 4.5.0/4.6.0 do not hang on any namespace in `girs/` — the defect was latent, with ~8× headroom, and generated output cannot have changed. The fix is still right; the ceiling was one deeply-generified declaration away.

## Also verified, fine as claimed

- Guard suite A/B: old generator → 4/6 fail in ~100 s with attributable messages (no hang), new → 6/6 pass.
- `Generator<string, never, unknown>`: removing the annotation reproduces `TS2322: Type 'string | void' is not assignable to type 'string'` at both `createGenericNameGenerator*` call sites; with it, 0 diagnostics. `next().value` types as `string` honestly — the loop cannot complete.
- Ignore-pattern coverage, both directions: a `debugger;` probe in `packages/reporter/lib/` (build output) is ignored (exit 0), the same probe in `packages/reporter/src/lib/` is reported (exit 1); no `lib/` directory exists at any other depth. Ten packages declare `outDir: lib` (the body says nine — prose only).
- The two `consistent-type-imports` warnings were fixed via `import type`, not suppressed; no disable comments anywhere in the diff.
- Remaining spread removals are alias-free by construction (`.map()`/`.filter()` allocate); the kept `[...this.interfaces]`/`[...this.generics]` spreads are still present. `GirModule.accept` snapshot is a pure hoist of the identical expression.
- Lockfile: `gjsify install --immutable` gates CI and passed; `vitest@^4.1.11`, `@types/node@^26.4.0`, `typescript@^6.0.3` were already in `gjsify-lock.json`'s requested set.

PR #464, opened 2026-09-08
fix(generator)!: enforce typed signals

sssxks:
## Problem

Generated GObject classes accept misspelled signal names and incompatible callbacks through
`string`/`any` fallback overloads, even when they have `SignalSignatures`. The typed `emit`
overload also drops the first declared argument: a signal taking `Variant | null` incorrectly
accepts an emission without that value.

## Changes

- Remove the permissive overloads from `connect`, `connect_after`, and `emit` by default, and
  preserve every declared signal argument in `emit`.
- Add the declaration-only `GObject.SignalMethods<Emitter, Signals>` helper. Custom classes
  refine their three inherited methods with `declare` fields, without runtime forwarding code.
  Merge `Gio.ListModel.SignalSignatures` into the two list model examples so `items-changed`
  connections and emissions stay checked.
- Document the full `registerClass` migration recipe, interface signals, dynamic GObject APIs,
  and the unchecked `notify::` property-name details.
- Extend `gobject-param-spec/strict-signals.ts` with a registered subclass that exercises custom
  and inherited signals, emission inside a class method, callback order, arguments, and
  disconnection in GJS. Check that further subclasses retain their callback emitter types.
- Compile consumers of freshly generated types to reject invalid names, callbacks, values, and
  argument counts. Compile and execute the README migration recipe as well.

BREAKING CHANGE: `connect`, `connect_after`, and `emit` reject unknown signal names and
incompatible arguments. Custom signals need typed signatures and method declarations, using
`GObject.SignalMethods`; truly dynamic calls can use the lower-level GObject signal APIs
explicitly. `notify::` details remain unchecked because of the inherited index signature.

## Validation

- `gjsify run build:types` and inspection of generated signal declarations
- `gjsify run build:examples`
- `gjsify run check`
- `gjsify run check:types`
- `gjsify run check:examples`
- `gjsify run test:tests`
- `gjsify run test:examples:cli`
- `gjsify workspace @ts-for-gir-test/strict-signals run test`
- GJS execution of `gio-2-list-model-example` and `virtual-interface-test`
- Formatting checks for changed sources and commitlint with the repository configuration

## LLM disclosure

Developed with Codex (initial implementation: GPT-6 Astra medium; review follow-up: GPT-6).

### DIFF
--- a/examples/di-needle/main.ts
+++ b/examples/di-needle/main.ts
@@ -208,6 +208,10 @@ class GObjectUserService extends GObject.Object {
 }
 
 // GObject-based Application service
+interface GObjectAppSignals extends Gio.Application.SignalSignatures {
+  "user-processed": (userName: string, userId: string) => void;
+}
+
 @injectable()
 class GObjectApp extends Gio.Application {
   static {
@@ -234,6 +238,27 @@ class GObjectApp extends Gio.Application {
   }
 
   private _appName = "User Management App";
+
+  override connect<K extends keyof GObjectAppSignals>(
+    signal: K,
+    callback: GObject.SignalCallback<this, GObjectAppSignals[K]>,
+  ): number {
+    return GObject.signal_connect(this, signal, callback);
+  }
+
+  override connect_after<K extends keyof GObjectAppSignals>(
+    signal: K,
+    callback: GObject.SignalCallback<this, GObjectAppSignals[K]>,
+  ): number {
+    return GObject.signal_connect_after(this, signal, callback);
+  }
+
+  override emit<K extends keyof GObjectAppSignals>(
+    signal: K,
+    ...args: GObject.GjsParameters<GObjectAppSignals[K]>
+  ): void {
+    GObject.signal_emit_by_name(this, signal, ...args);
+  }
 
   constructor(
     public readonly userService = inject(GObjectUserService),
--- a/examples/di-needle/main.ts
+++ b/examples/di-needle/main.ts
@@ -239,26 +239,9 @@ class GObjectApp extends Gio.Application {
 
   private _appName = "User Management App";
 
-  override connect<K extends keyof GObjectAppSignals>(
-    signal: K,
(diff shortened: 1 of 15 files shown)


sssxks:
@JumpLink  The reason for this change is we already have a mechanism for untyped signal connect, `GObject.signal_connect()`, but we still allow an untyped overload of `actor.connect()`. i wonder if you are comfortable with breaking changes. if so, i will remove the strict toggle and make it default.

JumpLink:
Hey @sssxks thank you! Yes go ahead, drop the toggle and make it default. `entry.connect("delete-txt", ...)`
compiling is exactly what the types exist to catch 👍

sssxks:
Updated as requested: the compatibility toggle is gone and strict signal overloads are now the default. The new fork SHA is waiting for an upstream Actions approval before CI can start; could a maintainer approve the pending runs when convenient?

JumpLink:
@sssxks  This looks good, and it already earned its keep: gtk-4-application declared `hello: (arg: string)` while the class actually registers `examplesig` with TYPE_INT. Both wrong, hidden by the fallback the whole time.

Two things before merge. commitlint is the only red job, two body lines are over 100 chars. And please mark it breaking, `fix(generator)!:` plus a BREAKING CHANGE line, otherwise release-it won't flag it and @girs consumers get it unannounced.

One inconsistency: di-needle and gtk-4-application keep type checking by declaring a SignalSignatures interface, but gio-2-list-model and virtual-interface-test drop to signal_emit_by_name and lose it:

```
	interface GjsListStoreSignals
	extends GObject.Object.SignalSignatures, Gio.ListModel.SignalSignatures {}
```

Also: none of the overriding classes override connect_after, so custom signals break there. Worth putting the full recipe in the README with connect, connect_after and emit, plus a note that `notify::` details stay unchecked because of the index signature. A test case with a registerClass subclass and a custom signal would be good too, that's the case everyone migrating will hit and the suite doesn't cover it.

JumpLink:
@sssxks thank you very much! I'll be releasing a new version soon.

